PR #8, opened 2016-11-11
Reworking directory structure

jeremyf:
## Beginning rearrangement of directories

@e645614d0f138cb0792ef2f01c15deb387fee05f

## Establishing directory structure for QA_tests

@12b3256fe4ec28e04467fec335712daaf327159d

## Switching to infer configuration

@06424447d3ce9a3cc3f6820ba0479473c5a96a5f

## Refactoring how spec helpers function

@d1d2ae42264c4ef8cafda15fe8340986494fc090

Removing triggers as those can be handled via a switch in the command:

```console
ENVIRONMENT=prod rspec
```


spec/microfilms/func_microfilms_spec.rb
```diff
@@ -37,6 +37,7 @@
     scenario 'Test 1: Enter text in search box' do
       visit '/'
       puts current_url
+
```
h-parekh: Is this a typo?

spec/spec_helper.rb
```diff
@@ -15,23 +15,25 @@
...
+RSpec.configure do |config|
+  config.include Capybara::DSL
 
-cnf = YAML::load_file(File.expand_path("../" + app_name + "/" + config_file, __dir__))
-trigger = YAML::load_file(File.expand_path("../" + app_name + "/" + trigger_file, __dir__))
-target_env = trigger['target']
+  # Ensuring that things run in a random order; This is a prefered mechanism
+  # as it helps identify temporal couplings
+  config.order = :random
+  Kernel.srand config.seed
```
h-parekh: Which order is being randomized? The order of setting application by host for each test suite?

spec/rbsc/functional/func_rbsc_spec.rb
```diff
@@ -1,4 +1,4 @@
-require 'spec_helper'
+require 'rbsc/rbsc_spec_helper'
```
h-parekh: I'm assuming the order of loading spec files when rspec test runs is in below order, correct?
spec/spec_helper.rb
spec/rbsc/rbsc_spec_helper.rb
spec/rbsc/functional/func_rbsc_spec.rb

spec/spec_helper.rb
h-parekh: So if we want to run only one application test suite (lets say spec/rbsc/functional/func_rbsc_spec.rb ) are you setting env variables for all apps?

jeremyf: The order in which each test is being run. It might jump from Curate to RBSC and back. If the scope is `rspec spec/curate` then all of the curate tests would be randomized.

spec/rbsc/functional/func_rbsc_spec.rb
jeremyf: Yes. I included `rbsc/rbsc_spec_helper.rb` as a hook for any additional tweaks related to RBSC specs.

spec/spec_helper.rb
```diff
@@ -33,6 +33,7 @@ def set_application_host_by_example(example)
   spec_helper_path = File.dirname(__FILE__)
   spec_path = example.metadata.fetch(:absolute_file_path)
   spec_sub_directory = spec_path.sub("#{spec_helper_path}/", '').split('/').first
+  # TODO: There may be a mix of environments that we want to test; How to handle this? The current assumption is test the same environment.
```
h-parekh: To keep it simple for now I think we can let the command line trigger specific envs: 
ENVIRONMENT=prod rspec 
ENVIRONMENT=pprd rspec  and so on.. 

That limits us to run tests sequentially though.

jeremyf: Yes. This gives the power to run sequential tests to the runner (without updating a repository file).

spec/spec_helper.rb
h-parekh: sweet.

jeremyf:
For each test, the capybara host is being set via the servers for the given
environment.

h-parekh:
Questions answered.

PR #12, opened 2016-11-16
refactor for using spec_helper

h-parekh:


h-parekh:
@jeremyf this refactor is same as the one I did for curate yesterday.

PR #18, opened 2016-11-23
Add QA tests for FAQ page

laritakr:
Filled in some basic tests for identifying that FAQ page is functioning.

### DIFF
--- a/spec/curate/pages/faq_page.rb
+++ b/spec/curate/pages/faq_page.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 module Curate
   module Pages
     # /faqs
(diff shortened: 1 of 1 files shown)


laritakr:
Modified structure for FAQ page tests. (Those still needing work are commented out.)

spec/curate/pages/faq_page.rb
```diff
@@ -1,12 +1,46 @@
...
   module Pages
-    # /about
+    # /faqs
     class FaqPage
       include Capybara::DSL
       include CapybaraErrorIntel::DSL
 
       def on_page?
-        # Fill out selectors etc.
+        status_response_ok? &&
```
jeremyf: Are you able to test that we are at the `/faqs` path?

spec/curate/pages/faq_page.rb
```diff
@@ -1,13 +1,55 @@
...
-        true
+        on_valid_url? &&
+          status_response_ok? &&
+          valid_page_content? &&
+          valid_page_navigation? &&
+          valid_external_page_links?
+      end
+
+      def on_valid_url?
+        current_url == (Capybara.app_host + 'faqs')
```
jeremyf: Use `File.join(Capybara.app_host, 'faqs')` as we don't know if there is a `/` between or not.

spec/curate/pages/faq_page.rb
```diff
@@ -6,12 +6,17 @@ class FaqPage
       include CapybaraErrorIntel::DSL
```
h-parekh: Can this line be placed in the 'RSpec.configure do |config|' block of spec_helper.rb  ?

spec/curate/pages/faq_page.rb
```diff
@@ -1,17 +1,47 @@
...
+        valid_page_navigation? &&
+        valid_external_page_links?
+      end
+
+      def status_response_ok?
+        self.status_code == 200
+      end
+
+      def valid_page_content?
+        self.has_content?('FAQ')
```
h-parekh: Let's put has.conten('FAQ') validation in a 'within' block like the other functions. 
helps avoid false positives for content validations. 

On a side note, is this a case sensitive validation?

spec/curate/pages/faq_page.rb
jeremyf: @h-parekh it _can_ be done, but it may involve more hoops to jump through. How about writing an issue in Github to explore/resolve this? Reference the issue in this pull request and we can move forward merging the code (and moving onto other tests) while capturing a possible refactor.

PR #20, opened 2016-11-28
Adding  TwP/logging

h-parekh:
Evaluating TwP/Logging

### DIFF
--- a/spec/curate/functional/func_curate_spec.rb
+++ b/spec/curate/functional/func_curate_spec.rb
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 require 'curate/curate_spec_helper'
 
+Logging.logger.root.appenders = Logging.appenders.stdout
+Logging.logger.root.level = :info
+
+log = Logging.logger['curate']
+
 feature 'User Browsing', js: true do
   #     scenario 'Test 1: Login to Curate with correct credentials' do
   #         visit '/'
@@ -93,7 +98,9 @@ feature 'User Browsing', js: true do
 
   require 'curate/pages/about_page'
   scenario 'Test 4: Go to About page' do
+    log.info RSpec.current_example.description
     visit '/'
+    log.info current_url
     click_on('About')
     print "Testing #{current_url}\n"
     about_page = Curate::Pages::AboutPage.new
--- a/Gemfile
+++ b/Gemfile
@@ -10,3 +10,4 @@ gem 'capybara_error_intel'
 gem 'rubocop', require: false
 gem 'rubocop-rspec'
 gem 'capybara-screenshot'
+gem 'logging'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -48,11 +48,16 @@ GEM
     json (1.8.3)
     launchy (2.4.3)
       addressable (~> 2.3)
+    little-plugger (1.1.4)
+    logging (2.1.0)
+      little-plugger (~> 1.1)
+      multi_json (~> 1.10)
     mime-types (3.1)
       mime-types-data (~> 3.2015)
     mime-types-data (3.2016.0521)
     mini_portile2 (2.1.0)
     minitest (5.9.1)
+    multi_json (1.12.1)
     net-scp (1.2.1)
       net-ssh (>= 2.6.5)
     net-ssh (3.2.0)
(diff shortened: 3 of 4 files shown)


spec/curate/functional/func_curate_spec.rb
```diff
@@ -1,6 +1,17 @@
 # frozen_string_literal: true
 require 'curate/curate_spec_helper'
 
+Logging.logger.root.appenders = Logging.appenders.stdout
+Logging.logger.root.level = :all
+
+log = Logging.logger['curate']
+
+log.debug 'Debug message'
```
jeremyf: Do these messages need to be here?

spec/curate/functional/func_curate_spec.rb
```diff
@@ -103,7 +114,10 @@
 
   require 'curate/pages/about_page'
   scenario 'Go to About page' do
+    #p RSpec.current_example.description
+    log.info RSpec.current_example.description
```
jeremyf: Move `log.info RSpec.current_example.description` into the [global before example hook](https://github.com/ndlib/QA_tests/blob/master/spec/spec_helper.rb#L33)

spec/curate/functional/func_curate_spec.rb
```diff
@@ -1,6 +1,17 @@
 # frozen_string_literal: true
 require 'curate/curate_spec_helper'
 
+Logging.logger.root.appenders = Logging.appenders.stdout
```
jeremyf: I believe these should be part of the `./spec/spec_helper.rb`

spec/curate/functional/func_curate_spec.rb
```diff
@@ -1,6 +1,17 @@
...
+Logging.logger.root.level = :all
+
+log = Logging.logger['curate']
```
jeremyf: I'm not following this command, why a custom log for 'curate'? Can this be moved into the [global before example hook](https://github.com/ndlib/QA_tests/blob/master/spec/spec_helper.rb#L33)

I also believe there might be unintended consequence by having these declared outside of the feature declarations. As defined, the `log` variable is a local variable within the Kernel. If we load three spec helpers throughout the rspec process, the `log` variable will be overwritten. If we end up running the spec in random order (which is advised) it may mean that the `log` variable is indeterminate; we may end up writing one test pass to the 'curate' log, then if we run again, to the `api` log.

spec/curate/functional/func_curate_spec.rb
```diff
@@ -84,14 +95,14 @@
 
   scenario 'Test 3: Enter text in search box' do
     visit '/'
-    puts current_url
+    #puts current_url
```
jeremyf: Just delete the lines. We have version control, commenting out code is more confusing than helpful.

spec/curate/functional/func_curate_spec.rb
```diff
@@ -84,14 +95,14 @@
...
     within('.homepage-search') do
       expect(page).to have_field 'catalog_search', type: 'search'
       expect(page).to have_button 'keyword-search-submit'
       fill_in('catalog_search', with: '')
       click_on('keyword-search-submit')
       # page.save_screenshot('curate_searchBox.png')
-      print "Clicked search\n"
+      #print "Clicked search\n"
```
jeremyf: Delete this line

spec/curate/functional/func_curate_spec.rb
```diff
@@ -103,7 +114,10 @@
 
   require 'curate/pages/about_page'
   scenario 'Go to About page' do
+    #p RSpec.current_example.description
```
jeremyf: Delete this line

spec/curate/functional/func_curate_spec.rb
h-parekh: Agree it shouldn't be in func_curate_spec. I'd like to have separate log files for each application so thinking if curate_spec_helper.rb would be a better place?

spec/curate/functional/func_curate_spec.rb
h-parekh: No these were experimental messages when I was trying to understand the scope of the different levels of logging.

spec/curate/functional/func_curate_spec.rb
h-parekh: will do

spec/curate/functional/func_curate_spec.rb
jeremyf: @h-parekh I believe we can set that in the global hook as well; based on the file path.

PR #24, opened 2016-11-30
Adding spec for help page

banurekha:


spec/curate/pages/help_page.rb
```diff
@@ -0,0 +1,32 @@
...
+      include CapybaraErrorIntel::DSL
+
+      def on_page?
+        on_valid_url? &&
+            status_response_ok? &&
+            modal_window_visible?
+      end
+
+      def on_valid_url?
+        current_url == File.join(Capybara.app_host)
```
jeremyf: I'm wondering about this. It looks like the behavior is a pop-up window. So we shouldn't assert any URL. And perhaps this should be renamed to `ModalHelpPage` class to draw attention to this being a modal page injected into another page.

spec/curate/pages/modal_help_page.rb
```diff
@@ -0,0 +1,27 @@
...
+        status_response_ok? &&
+        modal_window_visible?
+      end
+
+      def status_response_ok?
+        status_code == 200
+      end
+
+      def modal_window_visible?
+        find('#ajax-modal').visible?
```
jeremyf: I believe each of these lines need to be joined via an `&&` operator.

jeremyf: My understanding with the code as written is that lines 19, 20, 21 will execute, but only line 22 will be the return value (true or false). Which means if line 19 is false, but line 22 is true, the result of the method will be true.

jeremyf: In further review, the `has_field?` method calls `has_selector?` (https://github.com/teamcapybara/capybara/blob/master/lib/capybara/node/matchers.rb#L39) which will run an assert style check. However the `visible?` method does not appear to run an actual assert style method.

jeremyf:
@banurekha I went ahead and wrapped this up so we can move to the next things.

PR #25, opened 2016-11-30
Dltp 699 lkr tests

laritakr:
Expand search page testing to include qualified work type searches originating from the home page.


spec/curate/pages/catalog_page.rb
```diff
@@ -6,8 +6,27 @@ class CatalogPage
...
+        article: "Articles & Publications",
+        dataset: "Datasets & Related Materials"
+      }.freeze
+      LOOKUP_CATEGORY_VALUE = {
+        thesis: "Doctoral Dissertation OR Master's Thesis",
+        article: "Article",
+        dataset: "Dataset"
+      }.freeze
+
+      def initialize(options)
```
jeremyf: Go ahead and use [Ruby Keyword args](https://robots.thoughtbot.com/ruby-2-keyword-arguments); We are running this with Ruby 2.3.x.

```ruby
def initializer(search_term: = nil, category: = nil)
  @search_term = search_term
  @category = category
end
```

spec/spec_helper.rb
```diff
@@ -22,6 +22,8 @@
 
 # set the save path used in capybara-screenshot
 Capybara.save_path = './tmp/screenshots'
+# Keep only the screenshots generated from the last failing test suite
+Capybara::Screenshot.prune_strategy = :keep_last_run
```
h-parekh: We should keep last 3 failures, it will be useful in iterative troubleshooting. Capybara::Screenshot.prune_strategy = { keep: 3 }

spec/curate/pages/catalog_page.rb
laritakr: done

spec/spec_helper.rb
laritakr: Keep 3 means it keeps 3 screenshots from prior run(s). There's no way of really knowing how many we want. I'd recommend that unless we can find a way to keep x number of runs, we just comment out the prune when testing a problem.

h-parekh:
I think we can't merge these changes until @jeremyf  approves for the changes he requested.

spec/spec_helper.rb
jeremyf: I'm looking for clarification. Does `Capybara::Screenshot.prune_strategy = :keep_last_run` mean that if we run `rspec` and encounter 5 errors we will have 5 screenshots, then if we run `rspec` again with 3 errors, we will have 3 screenshots? Or does it mean if we run `rspec` and encounter 5 errors, only the last error encountered will have a screenshot?

jeremyf: Space is cheap, I'd recommend we keep at last the last 20 suite failures.

jeremyf: Let's not worry about this for the current pull request. We need to roll our own (but not 🚬 ) for run identifier storage.

PR #26, opened 2016-12-01
Test static asset

banurekha:
## Adding module to test network traffic on given url

@a7636c8bb05df5e9a5e872a1197e76612c9821a4

## fixing code review comments and running rubocop

@a28bfd51601f88f3ad40175dbba72eb774980bd8

## Adding after hook in spec_helper to verify static assets are loaded

@e7b439af227db0025b40825e0a953c9bac820a02

Adding module to verify the assets and include those test in after hook

## updating the assert statement

@03720d2a8845ffffe8dc53f0653ac9aeeb986aad


spec/spec_support/verify_network.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+module VerifyNetwork
+  module_function
+
+  def verify_network_traffic(page, object)
```
jeremyf: `object` is a generic and somewhat confusing variable name. I'm wondering about `test_handler` or `test_environment` as a variable name?

PR #30, opened 2016-12-05
Cap config updated to run 'bundle'

h-parekh:


config/deploy.rb
```diff
@@ -1,42 +1,14 @@
...
-# set :format_options, command_output: true, log_file: 'log/capistrano.log', color: :auto, truncate: :auto
-
-# Default value for :pty is false
-# set :pty, true
-
-# Default value for :linked_files is []
-# append :linked_files, 'config/database.yml', 'config/secrets.yml'
-
-# Default value for linked_dirs is []
-# append :linked_dirs, 'log', 'tmp/pids', 'tmp/cache', 'tmp/sockets', 'public/system'
```
jeremyf: @h-parekh Did you want to link the log directory? This would allow us to preserve logs from deploy to deploy?

h-parekh: Yes this is a good idea. I'll update the cap config once I get through the logging implementation.

jeremyf: Excellent.

PR #35, opened 2016-12-07
Adding timestamp

h-parekh:
Committing these to avoid merge conflict grief.

### DIFF
--- a/spec/curate/functional/func_curate_spec.rb
+++ b/spec/curate/functional/func_curate_spec.rb
@@ -110,14 +110,14 @@ feature 'User Browsing', js: true do
   end
 
   require 'curate/pages/faq_page'
-  scenario 'Test: Go to FAQ page' do
+  scenario 'Test start: Go to FAQ page' do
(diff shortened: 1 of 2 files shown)


spec/curate/functional/func_curate_spec.rb
```diff
@@ -1,10 +1,14 @@
 # frozen_string_literal: true
 require 'curate/curate_spec_helper'
 
-Logging.logger.root.appenders = Logging.appenders.stdout
-Logging.logger.root.level = :info
+#Logging.logger.root.appenders = Logging.appenders.stdout
```
jeremyf: Remove commented out code; We have version control. Commented out code can be confusing when someone comes to it later. Why is that commented out? What was the trigger for commenting out the code.

h-parekh: Sorry, my bad. This was work in progress and I wanted to commit the other changes. I've removed those comments in the upcoming commit.

Note to self: Never commit with commented code.

PR #37, opened 2016-12-08
Pushing curate page requirements to helper

jeremyf:
Instead of the explicit require, assume we will use all of the pages
throughout the spec. This removes some mental overhead.


h-parekh:
@jeremyf  you have merge conflict.

PR #38, opened 2016-12-08
Facet navigation

banurekha:


spec/curate/functional/func_curate_spec.rb
```diff
@@ -148,3 +148,91 @@
...
+    expect(page).to have_selector('#ajax-modal', visible: true)
+    expect(page).to have_content('Collection')
+    print "Clicked  Department or Unit \n"
+    within('#ajax-modal') do
+      find('.close').click
+    end
+    expect(page).not_to have_selector("#ajax-modal")
+  end
+
+  scenario 'Type of Work' do
```
jeremyf: It looks like these scenarios are very similar. Can you create an iterating loop:

```ruby
["Type_of_Work", "Creator"].each do |facet_name|
  scenario facet_name do
    visit "/"
    click_on("Search")
    expect(page).not_to have_css("#collapse_#{facet_name}.in.collapse")
    find(%(a[data-target="#collapse_#{facet_name}"])).click
    expect(page).to have_css("#collapse_#{facet_name}.in.collapse")
  end
end
```

PR #41, opened 2016-12-08
Adding auto-logging in standard format

jeremyf:
* click, visit, and submit events in a browser
* network traffic
* starting and stopping

This helps reduce the chatter of logging within a scenario's body
body. This improves legibility of the spec. It also improves the
predictability of formatting.


spec/spec_helper.rb
```diff
@@ -12,9 +12,10 @@
 require 'capybara-screenshot/rspec'
 require 'logging'
 require 'rspec/logging_helper'
-require 'spec_support/verify_network'
+require 'spec_support/example_logging'
```
h-parekh: Can we do a 'require all files in spec_support' similar to https://github.com/ndlib/QA_tests/blob/master/spec/curate/curate_spec_helper.rb#L3

jeremyf: @h-parekh That's a good suggestion. I'll make add a commit

spec/spec_support/example_logging.rb
```diff
@@ -0,0 +1,224 @@
...
+  # @note Call this method no more than once per spec
+  def self.instantiate_all_loggers!(config: ENV)
+    raise "You already called instantiate_all_loggers!" if @called
+    @called = true
+    Logging.appenders.stdout(layout: Logging.layouts.pattern(format_as: :json))
+    entries = Dir.glob(File.expand_path('../../*', __FILE__))
+    entries.each do |entry|
+      logger = Logging.logger[entry.split('/').last]
+      logger.add_appenders('stdout')
+      logger.level = config.fetch('LOG_LEVEL', :info).to_sym
```
h-parekh: Are you setting LOG_LEVEL to info if no values are passed? If yes, then that's already happening at https://github.com/ndlib/QA_tests/pull/41/files#diff-2f39a8464db2dbee6055fc09db08e61bR12

bin/run_tests
```diff
@@ -0,0 +1,47 @@
...
+  $stdout.puts ""
+  $stdout.puts "$ ./bin/#{File.basename(__FILE__)} [path/to/spec]"
+  $stdout.puts ""
+  $stdout.puts "To run all of the tests:"
+  $stdout.puts "  $ ./bin/run_tests"
+  $stdout.puts ""
+  $stdout.puts "To run some of the tests:"
+  $stdout.puts "  $ ./bin/run_tests path/to/spec.rb"
+  $stdout.puts ""
+  $stdout.puts "You may also pass any options of the options of rspec to this script"
```
h-parekh: Should we explicitly enlist all available options for ENVIRONMENT and LOG_LEVEL? Especially for log_level I think the order is important to know when someone runs a test. I also tested with LOG_LEVEL=all and it worked like a charm.

jeremyf: Do we know the ENVIRONMENT options? Those are convention based; The LOG_LEVEL we can take a stab at.

README.md
```diff
@@ -93,25 +93,33 @@ As such, you should run the specs so that our tests are not writing log informat
 
 (TODO: Provide scripts to run tests)
```
h-parekh: TODO complete if you're referring to /bin/run_tests.

bin/run_tests
h-parekh: https://github.com/ndlib/QA_tests/blob/master/spec/curate/curate_config.yml

jeremyf: @h-parekh But not all applications we test have staging6, staging8, or staging9. There is a larger question then…what is the expected behavior of running the following:

```console
ENVIRONMENT=staging6 ./bin/run_tests
```

Should we run just tests that have staging6 as an environment? Should we fail?

spec/spec_support/example_logging.rb
```diff
@@ -0,0 +1,226 @@
...
+require 'active_support/core_ext/module/delegation'
+
+# This module provides mechanisms for consolidating logging
+#
+# It is intended to be included in the RSpec example context so as to expose
+# the @current_logger instance variable
+module ExampleLogging
+  DEFAULT_ENVIRONMENT = 'prod'
+  DEFAULT_LOG_LEVEL = 'info'
+  AVAILABLE_LOG_LEVELS = %w(debug info warning error fatal)
```
jeremyf: Good catch.

spec/spec_support/example_logging.rb
```diff
@@ -0,0 +1,226 @@
+# frozen_string_literal: true
+require 'active_support/core_ext/module/delegation'
+
+# This module provides mechanisms for consolidating logging
+#
+# It is intended to be included in the RSpec example context so as to expose
+# the @current_logger instance variable
+module ExampleLogging
+  DEFAULT_ENVIRONMENT = 'prod'
```
h-parekh: We may need to rework the default env logic once we get the login and 'upload-data' test scenarios setup. Otherwise tests will fail against prod.

jeremyf: @h-parekh Yes. The default assumption starts to change once we look at how we leverage the metadata tags for each of the tests. For now, I believe this sticks as we don't know what that next bit looks like.

bin/run_tests
h-parekh: Test should abort and display available options for ENVIRONMENT variable based on appName_config.yml for the test case being run.

spec/spec_support/example_logging.rb
h-parekh: Yep, lets stick with this for now. Just wanted to put it out there since it came to my mind. 
I'm going to keep this on my radar to review once we get to non-prod specific tests.

bin/run_tests
jeremyf: @h-parekh And if we are running all of the specs for all of the applications? The current construction of the test runner (just like the raw rspec predecessor) has no knowledge of the applications that can be tested.

I don't disagree that this would be helpful, I believe it is beyond the scope of this change (as you are now implying that we will always launch the testing process against a single application).

spec/spec_support/example_logging.rb
```diff
@@ -0,0 +1,226 @@
+# frozen_string_literal: true
+require 'active_support/core_ext/module/delegation'
```
h-parekh: This looks like an extension, does it need any additional setup on local/remote? Going by the prefix 'core' I'm thinking it should already be there, but just wanted to confirm.

jeremyf: I believe that is a no longer needed line. Will remove.

But to your specific question, it is a core extension, however it is not automatically required in our setup. Rails would normally automatically require that file.

spec/spec_support/example_logging.rb
```diff
@@ -0,0 +1,226 @@
...
+              error(context: "verifying_network_traffic", url: response.url, status_code: response.status)
+            else
+              debug(context: "verifying_network_traffic", url: response.url, status_code: response.status)
+            end
+          end
+          test_handler.expect(failed_resources).to test_handler.be_empty, build_failed_messages_for(failed_resources)
+        end
+      end
+
+      def build_failed_messages_for(failed_resources)
```
h-parekh: At what level are failed_resources being logged? Or those logs are outside the logging framework?

jeremyf: There are a few logging outputs:

1) Each application has a custom logger
2) RSpec is logging to custom locations (as per 'rspec/logging_helper' from the Logging gem)

Any "expect" calls will log to the RSpec logging. It looks like that goes to STDOUT for the time being. The question we'll have is how do you want these intermixed? I'd like to propose that as work to be done outside of this pull request.

h-parekh: Let's whiteboard this next week in office.

PR #53, opened 2017-01-04
Creating run ID for tracking test results

h-parekh:
Using the before(:suite) block to create a run ID based on datetime. 
After block is WIP

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -43,6 +43,10 @@ RSpec.configure do |config|
   config.order = :random
   Kernel.srand config.seed
 
+  config.before(:suite) do
+    RunIdentifier.set
+  end
+
   config.before(:example) do |rspec_example|
     @current_logger = ExampleLogging.start(example: rspec_example, config: ENV, test_handler: self)
   end
--- a/spec/spec_support/example_logging.rb
+++ b/spec/spec_support/example_logging.rb
@@ -200,11 +200,11 @@ module ExampleLogging
         message = ""
         kwargs.each { |key, value| message += %(#{key}: #{value.inspect}\t) }
         if block_given?
-          @current_logger.public_send(severity, %(test_type: #{test_type}\t context: "BEGIN #{context}\t#{message}).strip)
+          @current_logger.public_send(severity, %(test_type: #{test_type}\t runID: #{RunIdentifier.get}\t context: "BEGIN #{context}\t#{message}).strip)
           yield
-          @current_logger.public_send(severity, %(test_type: #{test_type}\t context: "END #{context}\t#{message}).strip)
+          @current_logger.public_send(severity, %(test_type: #{test_type}\t runID: #{RunIdentifier.get}\t context: "END #{context}\t#{message}).strip)
         else
-          @current_logger.public_send(severity, %(test_type: #{test_type}\t context: "#{context}"\t#{message}).strip)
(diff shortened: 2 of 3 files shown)


spec/spec_support/test_runtime.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+module TestRuntime
+  def self.create_run_id
+    d = DateTime.now
```
jeremyf: I prefer more descriptive variable names than `d`. It doesn't describe what it is. Consider calling it `now`.

spec/spec_support/test_runtime.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+module TestRuntime
```
jeremyf: The module name is a bit of a misnomer; You are using it for as a means of setting a `RunIdentifier`.

How about this interface:

```ruby
module RunIdentifier
  def self.set
    # Set the identifier
  end
  def self.get
    # Set the identifier
  end
end
```

spec/spec_support/test_runtime.rb
```diff
@@ -0,0 +1,8 @@
...
+  def self.create_run_id
+    d = DateTime.now
+    $runId = d.strftime("%m%d%Y_%H%M%S.%L")
```
jeremyf: This module is encapsulating the setting of a global constant. However, you are not encapsulating the reading of that global constant. Unless there is no way around a global variable, opt not to use those variables.

```ruby
module RunIdentifier
  def self.set
    @run_identifier = DateTime.now.strftime("%m%d%Y_%H%M%S.%L")
  end
  def self.get
    @run_identifier
  end
end
```

The above solution sets a module instance variable, that module instance variable will persist so long as the module persists (which is for the whole duration of the tests).

spec/spec_support/example_logging.rb
```diff
@@ -200,11 +200,11 @@ def log(severity:, context:, **kwargs)
         message = ""
         kwargs.each { |key, value| message += %(#{key}: #{value.inspect}\t) }
         if block_given?
-          @current_logger.public_send(severity, %(test_type: #{test_type}\t context: "BEGIN #{context}\t#{message}).strip)
+          @current_logger.public_send(severity, %(test_type: #{test_type}\t runID: #{$runId}\t context: "BEGIN #{context}\t#{message}).strip)
           yield
-          @current_logger.public_send(severity, %(test_type: #{test_type}\t context: "END #{context}\t#{message}).strip)
+          @current_logger.public_send(severity, %(test_type: #{test_type}\t runID: #{$runId}\t context: "END #{context}\t#{message}).strip)
```
jeremyf: Instead of accessing `$runId` us an accessor on the module in which the `$runId` is set. This encapsulation allows for future changes.

spec/spec_helper.rb
```diff
@@ -37,12 +37,17 @@
   config.include RSpec::LoggingHelper
   config.include ExampleLogging
   config.include ExampleLogging::CapybaraInjection
+  config.include TestRuntime
```
jeremyf: This is not necessary. The `config.include` method call is only necessary when we want to append the methods of the given module into the `it` blocks of the specs. As the `TestRuntime` module does not have methods that we need to access within the tests, the include is not necessary.

In addition, all module functions of a module – those methods defined in a module with `self.` – are not included in the methods that would be appended.

jeremyf:
And last but not least, can you squash the commits?

PR #76, opened 2017-03-07
Hp cloudwatch events

h-parekh:
PR to send custom JSON output results from Rspec as cloudwatch events.

### DIFF
--- a/spec/spec_support/cloudwatch_event_handler.rb
+++ b/spec/spec_support/cloudwatch_event_handler.rb
@@ -4,11 +4,15 @@ require 'aws-sdk'
 module CloudwatchEventHandler
   def self.set_aws_config
     target_aws_account = 'testlibnd'
-
     user_home_dir = File.expand_path('~')
     aws_env_config_file = YAML.load_file(File.join("#{user_home_dir}", 'test_data/QA/aws_config.yaml'))
-    require 'byebug'; debugger
     Aws.config.update({region: aws_env_config_file.fetch("#{target_aws_account}").fetch("default_region_name")})
     Aws.config.update({credentials: Aws::Credentials.new(aws_env_config_file.fetch("#{target_aws_account}").fetch("aws_access_key_id"), aws_env_config_file.fetch("#{target_aws_account}").fetch("aws_secret_access_key"))})
+  end
+
+  def self.report_json_result(output_hash:)
+    client = Aws::CloudWatchEvents::Client.new
+    rspec_result_event = {entries:[{"source": "testcontroller01","detail_type": "rspec_result","detail": output_hash.to_json}]}
+    client.put_events(rspec_result_event)
   end
 end
--- a/spec/spec_support/cloudwatch_event_handler.rb
+++ b/spec/spec_support/cloudwatch_event_handler.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+require 'aws-sdk'
+
+module CloudwatchEventHandler
+  def self.set_aws_config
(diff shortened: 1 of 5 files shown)


spec/spec_support/cloudwatch_event_handler.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+require 'aws-sdk'
+
+module CloudwatchEventHandler
+  def self.set_aws_config
```
jeremyf: @h-parekh **ping**

spec/spec_support/json_output_formatter.rb
```diff
@@ -7,8 +7,13 @@ class JsonOutputFormatter < RSpec::Core::Formatters::JsonFormatter
   # https://github.com/rspec/rspec-core/blob/master/lib/rspec/core/formatters/json_formatter.rb
   RSpec::Core::Formatters.register self
 
-  def stop(notification)
+  def close(notification)
```
h-parekh: Both methods receive rspec output notification at the end of the suite, but the close method also includes 'summary' and 'summary_line' which can serve as metrics in cloudwatch.

Example:
summary":{"duration":4.592754,"example_count":1,"failure_count":0,"pending_count":0},"summary_line":"1 example, 0 failures"}

spec/spec_support/cloudwatch_event_handler.rb
h-parekh: I want to keep the secrets in corpFS along with all other QA test data (includes test accounts). Also these secrets will change very rarely and will be the same when tests are running on local or testcontroller01.

Are you suggesting to use Figaro for combining test user account credentials and AWS credentials in the same file?

jeremyf: Yes. I'm suggesting all secrets be managed by Figaro and we declare as part of the boot what keys are required.

h-parekh: I like the part where Figaro allows you to require keys at boot, helps prevent runtime errors due to configuration. 

From my understanding so far I'm not sure if it's going to be easy to use Figaro in this case beacuse:
1. "Figaro parses a Git-ignored YAML file in your application and loads its values into ENV." That means I have to place all secrets in Github instead of corpFS?

2.  "ENV is a simple key/value store. Deeply nested configuration structures are not possible." Our user account credentials file is a CSV that needs username, password and passcode all together to make a login work. I'm not sure how to fit that easily into a single key:value pairs.

I may be missing something here, thoughts?

jeremyf: 1. Secrets would be on CorpFS. You would have dummy secrets locally that are committed to Github. When the test controller application is deployed, those secrets are copied (as per how Sipity, Curate, and Hesburgh Matrics now deploy)

2. For account credentials, we could also store them on CorpFS and copy them at deploy. If we want to use Figaro for the test user accounts, we'd need to flatten the CSV.

h-parekh: Here's what I think are the benefits of using this approach:
* Provides a consistent way to access any key within the app
* Avoids using multiple YAML.load_file statements across code
* Since it's going to copy files at deploy, it may capture any disk read/write issues instead of at runtime

jeremyf: @h-parekh Avoiding multiple loads (and the adhoc nature of that loading) is a massive improvement for me.

jeremyf:
I'm merging this so we can see the information. I still want to move towards Figaro

PR #79, opened 2017-03-08
Web crawler site verification

NicHaydel95:


spec/crawler/crawl/web_crawler_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      encoded_url = URI.encode(url)
+      doc = Nokogiri::HTML(open(encoded_url))
+      root = url
+      links = doc.css('a')
+      hrefs = links.map { |link| link.attribute('href').to_s }.uniq.sort.delete_if { |href| href.empty? }
+      for link in hrefs do
+        if link.start_with?('/')
+          link = root + link
+        end
+        if link.include?("#") || link == " " || link =~ /mailto:(.*)/ || link.nil?
```
jeremyf: Instead of an empty `if` with a meaningful `else`, prefer `next if link.include?("#") || link == " " || link =~ /mailto:(.*)/ || link.nil?`

Remove the `link.nil?` check as it is not necessary (if link is nil line 17 will fail)

spec/crawler/crawl/web_crawler_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+        else
+          uri = URI.parse(link)
+          begin
+            response = Net::HTTP.get_response(uri)
+            if response.code.to_i >= 400 && response.code.to_i <= 599
+              current_logger.info(context: "crawler", url: link, status_code: response.code)
+            end
+          # Captures known errors from links to mailto: and tel:
+          # ex. mailto:email@nd.edu
+          rescue SystemCallError
```
jeremyf: Instead of rescuing mailto: and tel: errors, skip those links in the above if imperative.

spec/crawler/crawl/web_crawler_spec.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+require 'crawler/crawler_spec_helper'
+require 'Nokogiri'
+require "net/http"
+require "uri"
+feature 'Link Checker' do
+  scenario 'Check Links' do
+    File.foreach(ENV["HOME"] + "/crawl_sites.txt") do |url|
+      url.delete!("\n")
```
jeremyf: Prefer `url.strip!` ensures we clean-up all tabs, new lines, and carriage returns.

spec/crawler/crawl/web_crawler_spec.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+require 'crawler/crawler_spec_helper'
+require 'Nokogiri'
+require "net/http"
+require "uri"
+feature 'Link Checker' do
+  scenario 'Check Links' do
+    File.foreach(ENV["HOME"] + "/crawl_sites.txt") do |url|
```
jeremyf: Rename to `root_url`; Later in the code you set `root = url`

spec/crawler/crawl/web_crawler_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      url.delete!("\n")
+      current_logger.info(context: "Begin verifying: #{url}")
+      encoded_url = URI.encode(url)
+      doc = Nokogiri::HTML(open(encoded_url))
+      root = url
+      links = doc.css('a')
+      hrefs = links.map { |link| link.attribute('href').to_s }.uniq.sort.delete_if { |href| href.empty? }
```
jeremyf: This algorithm loops through links 4 times. Instead prefer something that loops once. I'm also wondering why the links need to be sorted?

```ruby
require 'set'
hrefs = Set.new
links.each do |link|
  href = link.attribute('href').to_s.strip
  hrefs << href unless href.empty?
end
```

spec/crawler/crawl/web_crawler_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      for link in hrefs do
```
jeremyf: When looping through an array, the element should be the singular version of the array's name (eg. `for href in hrefs do`). This is especially important given you have `links` on line 14 and someone might assume that you are looping through those.

spec/crawler/crawl/web_crawler_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+        if link.start_with?('/')
+          link = root + link
+        end
+        next if link.include?("#") || link == " " || link =~ /mailto:(.*)/ || link.nil? || link=~ /tel:(.*)/
+        uri = URI.parse(link)
+        response = Net::HTTP.get_response(uri)
+        if response.code.to_i >= 400 && response.code.to_i <= 599
+          current_logger.info(context: "crawler", url: link, status_code: response.code)
```
jeremyf: @h-parekh I believe this should be an error?

spec/crawler/crawl/web_crawler_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      encoded_url = URI.encode(url)
+      doc = Nokogiri::HTML(open(encoded_url))
+      root = url
+      links = doc.css('a')
+      hrefs = links.map { |link| link.attribute('href').to_s }.uniq.sort.delete_if { |href| href.empty? }
+      for link in hrefs do
+        if link.start_with?('/')
+          link = root + link
+        end
+        next if link.include?("#") || link == " " || link =~ /mailto:(.*)/ || link.nil? || link=~ /tel:(.*)/
```
jeremyf: You already tested that link is not nil? (based on the above href.empty?). Likewise you can get rid of `link == " "` test (if you strip the href).

I don't believe you should be skipping all links that `include?('#')`; We may be anchoring to a place on another page. Instead I believe you want to go to the next link if the link matches `link =~ /\A#/`.

For the mailto: and tel: regular expressions, you don't need to capture (e.g. `(.*)`) the value after the `:`. You should be able to use `/mailto:.*/` and `/tel:.*/` or simply `/mailto:/` and `/tel:/`; Capturing values from regular expressions slows the evaluation down (a bit).

spec/crawler/crawl/web_crawler_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      end
+      for href in hrefs do
+        if href.start_with?('/')
+          href = root_url + href
+        end
+        next if href =~ /\A#/ || href =~ /mailto:/ || href =~ /tel:/
+        uri = URI.parse(href)
+        response = Net::HTTP.get_response(uri)
+        if response.code.to_i >= 400 && response.code.to_i <= 599
+          current_logger.info(context: "crawler", url: href, status_code: response.code)
```
jeremyf: @h-parekh Are we wanting to log these as info? Or should they be error? Should was assert a failure?

As written, this spec will pass regardless of any broken links.

h-parekh: I'm thinking of this as an auditing tool instead of our traditional QA_tests. I don't want the test to fail because if a link's broken the scenario will stop and won't test all remaining links.

PR #82, opened 2017-03-15
Adding ability to skip cloudwatch

jeremyf:


bin/run_tests
```diff
@@ -9,10 +9,11 @@ require File.expand_path('../../spec/spec_support/example_logging_constants', __
 #
 # *****************************************************************************
 
-CONFIG_KEYS = [:ENVIRONMENT, :LOG_LEVEL].freeze
+CONFIG_KEYS = [:ENVIRONMENT, :LOG_LEVEL, :SKIP_CLOUDWATCH].freeze
```
h-parekh: SKIP_CLOUDWATCH will only be passed as a env variable without any value assigned to it, correct? May be add a stdout.puts to explain usage?

jeremyf: @h-parekh I added notes in the help for this script (e.g. `./bin/run_tests -h` for information)

jeremyf:
Not quite. ENV values are converted to strings. And in Ruby, all strings
evaluate as true. The SKIP_CLOUDWATCH is not something we would set via
Jenkins.

Jeremy Friesen

PR #89, opened 2017-03-17
Renaming webRenovation files and references

h-parekh:
MacOS filesystem is case insensitive and Linux FS is case sensitive.

```console
require 'webrenovation/webrenovation_spec_helper'
```
This statement will work in MacOS for a folder structure like
'spec/webRenovation/webrenovation_spec_helper.rb', but fails on linux

I'm renaming files and their references to match up
with 'webRenovation'


jeremyf:
Thanks for the context of why you were renaming (linux case sensitivity)

PR #91, opened 2017-03-23
Modifications to Swagger handler

h-parekh:
* Adding support for private git repositories
* Parameterizing swagger definition URL based on repository URL from a config file

### DIFF
--- a/spec/spec_support/swagger_handler.rb
+++ b/spec/spec_support/swagger_handler.rb
@@ -31,9 +31,17 @@
     end
 
     def download_and_load_definition!
-      url_to_swagger_def = File.join('https://raw.githubusercontent.com/', project_user_name, project_repository_name, 'master', 'definitions', 'swagger.yaml')
-      swagger_yaml = open(url_to_swagger_def).read
+      url_to_swagger_def = File.join('https://raw.githubusercontent.com/', project_user_name, project_repository_name, 'master', 'definitions', 'swagger.yml')
+      read_git_access_token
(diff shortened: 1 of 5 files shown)


spec/spec_support/swagger_handler.rb
```diff
@@ -20,20 +22,27 @@ def operations(&block)
 
     private
 
-    def project_user_name
-      'ndlib'
-    end
-
-    def project_repository_name
-      'test_api'
+    def   read_repo_config
```
jeremyf: one space between def and the method name.

jeremyf: I recommend that you set all instance variables as part of the initialize process. Then, you can call the other methods that use the instance variables at a later point. This helps clean up and separate concerns of the code.

spec/viceroy/integration/int_viceroy_spec.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+require 'viceroy/viceroy_spec_helper'
+
+describe 'API tests using swagger definition' do
+  SwaggerHandler.operations(for_file_path: __FILE__).each do |operation|
```
jeremyf: I'm fine with this, but now that we have two use cases of Swagger operations, we should look to remove the duplication of knowledge. It gets a bit tricky, but I can help on that behavior.

spec/spec_support/swagger_handler.rb
h-parekh: I moved 'read_repo_config' and 'read_git_access_token' under initialize.

PR #96, opened 2017-04-07
Removing webkit

h-parekh:
We've found out that the poltergeist driver works well for ReactJS
applications as well.
I'm taking this out of the testing framework to simplify
workstation setup time for other developers. QT5 setup is very cumbersome for MacOS.


jeremyf:
Happy Friday Dance! Removing build dependencies!

PR #101, opened 2017-04-17
Screenshots error

h-parekh:
* Addressing issue with tmp/screenshots directory not being found.
* Adding documentation
* Directory cleanup mechanism used to delete latest dir instead of oldest one

### DIFF
--- a/spec/spec_support/test_runtime.rb
+++ b/spec/spec_support/test_runtime.rb
@@ -21,9 +21,9 @@ module RunIdentifier
     FileUtils.mkdir_p(screenshots_root)
     Dir.chdir(screenshots_root)
     screenshots_directories = Dir.glob('*').select { |f| File.directory? f } # Returns a list of all screenshots directories in screenshots_root
-    screenshots_directories.each_with_index { |dir, index|
+    screenshots_directories.reverse.each_with_index { |dir, index|
       if index>=runs
-        FileUtils.rm_rf(screenshots_directories.first)
+        FileUtils.rm_rf(dir)
       end
     }
 
--- a/spec/spec_support/test_runtime.rb
+++ b/spec/spec_support/test_runtime.rb
@@ -2,7 +2,6 @@
 require 'fileutils'
 
 module RunIdentifier
-
   def self.set
     @run_identifier = DateTime.now.strftime("%Y-%m-%dT%H:%M:%S.%L-05:00")
   end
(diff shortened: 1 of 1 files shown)


spec/spec_support/test_runtime.rb
```diff
@@ -11,24 +10,26 @@ def self.get
...
     current_working_directory = Dir.pwd
+    FileUtils.mkdir_p(screenshots_root)
     Dir.chdir(screenshots_root)
-    screenshots_directories = Dir.glob('*').select {|f| File.directory? f} # Returns a list of all screenshots directories in screenshots_root
-    screenshots_directories.each_with_index{|dir, index|
+    screenshots_directories = Dir.glob('*').select { |f| File.directory? f } # Returns a list of all screenshots directories in screenshots_root
+    screenshots_directories.each_with_index { |dir, index|
       if index>=runs
-        FileUtils.rm_rf(dir)
+        FileUtils.rm_rf(screenshots_directories.first)
```
jeremyf: Why did you change this? Is it deleting the later runs?

I believe that you need to keep `FileUtils.rm_rf(dir)` instead of getting rid of the first one. Your change will likely cause an error if we have 5 entries and are told to keep only 2 (as per the `runs:` keyword parameter). The first time through, all will be well. The second time through, it will attempt to delete the already deleted first directory.

h-parekh: Yes FileUtils.rm_rf(dir) deletes the directory whose index is 6 in screenshots_directories. That directory contains screenshots from the latest run. 

I want to keep the latest screenshots for comparing consecutive test runs.

spec/spec_support/test_runtime.rb
```diff
@@ -11,24 +10,26 @@ def self.get
...
+  # @return [String]
   def self.get_screenshots_save_path(runs: 5)
     screenshots_root = 'tmp/screenshots/'
     current_working_directory = Dir.pwd
+    FileUtils.mkdir_p(screenshots_root)
     Dir.chdir(screenshots_root)
-    screenshots_directories = Dir.glob('*').select {|f| File.directory? f} # Returns a list of all screenshots directories in screenshots_root
-    screenshots_directories.each_with_index{|dir, index|
+    screenshots_directories = Dir.glob('*').select { |f| File.directory? f } # Returns a list of all screenshots directories in screenshots_root
+    screenshots_directories.each_with_index { |dir, index|
```
jeremyf: If we change this to `screenshots_directories.reverse.each_with_index { |dir, index|` and restore `FileUtils.rm_rf(dir)` this will behave as intended.

PR #103, opened 2017-04-19
Pretty printing JSON output for better readability

h-parekh:
The 'close' method of parent class simply spits out the entire JSON
output which is very hard to read. I'm removing 'super' and explicitly
using JSON.pretty_generate instead.

### DIFF
--- a/spec/spec_support/json_output_formatter.rb
+++ b/spec/spec_support/json_output_formatter.rb
@@ -9,8 +9,8 @@ class JsonOutputFormatter < RSpec::Core::Formatters::JsonFormatter
   RSpec::Core::Formatters.register self
 
   def close(notification)
-    super
     @output_hash[:runID] ||= "#{RunIdentifier.get}"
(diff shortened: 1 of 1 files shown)


spec/spec_support/json_output_formatter.rb
```diff
@@ -9,8 +9,8 @@ class JsonOutputFormatter < RSpec::Core::Formatters::JsonFormatter
   RSpec::Core::Formatters.register self
 
   def close(notification)
-    super
```
jeremyf: Why are you removing `super` usage?

h-parekh: super usage calls the [close method of parent class](https://github.com/rspec/rspec-core/blob/master/lib/rspec/core/formatters/json_formatter.rb#L55), which is simply printing json to console:
output.write @output_hash.to_json

``` ruby      
  def close(_notification)
      output.write @output_hash.to_json
  end
```

PR #104, opened 2017-04-21
WR-301

danthewolfe:
As a developer and QA specialist, I want to write validations for the
main elements of homepage.

### DIFF
--- a/spec/webRenovation/pages/home_page.rb
+++ b/spec/webRenovation/pages/home_page.rb
@@ -7,17 +7,58 @@ module WebRenovation
       include CapybaraErrorIntel::DSL
 
       def on_page?
-        super && on_valid_url?
-
-        within('#wrapper') do
-          find_link('University of Notre Dame').visible?
-        end
+        super &&
+          on_valid_url? &&
+          has_nd_branding? &&
+          searchbox_valid? &&
+          has_service_boxes? &&
+          has_news? &&
+          has_events? &&
+          has_hours?
       end
 
       def on_valid_url?
         current_url == Capybara.app_host || current_url == (Capybara.app_host + "/")
       end
 
+      def has_nd_branding?
+        within('#wrapper') do
+          find_link('University of Notre Dame').visible?
+        end
+      end
+
+      def searchbox_valid?
+        find_by_id('searchAppliance') &&
+          find_button('Search')
+      end
+
+      def has_service_boxes?
+        # Reserves
+        find_link(:href => 'https://reserves.library.nd.edu', :title => 'Course Reserves')
+        # ILL
+        find_link(:href => 'https://nd.illiad.oclc.org/illiad/IND/illiad.dll', :title => 'Interlibrary Loan and Document Delivery')
+        # Reserve Room
+        find_link(:href => 'http://nd.libcal.com/#s-lc-box-2749-container-tab1', :title => 'Reserve a Room')
+      end
+
+      def has_news?
+        find('h3', text: 'News') &&
+          all('.news-card', between: 1..3)
+      end
+
+      def has_events?
+        find('h3', text: 'Events') &&
+          all('.event-card', between: 1..3)
+      end
+
+      def has_hours?
+        # Hours display mid page
+        all(:css, 'a[href="/page/hours/"]')[0].text != ''
+        all(:css, 'a[href="/page/hours/"]')[0].text != nil
+
+        # Hours button in footer
+        find_link(:href => '/page/hours/', :class =>'hours')
+      end
     end
   end
 end
--- a/spec/webRenovation/pages/home_page.rb
+++ b/spec/webRenovation/pages/home_page.rb
@@ -32,14 +32,13 @@ module WebRenovation
       end
 
       def has_service_boxes?
-
         within('.row.services') do
           # Reserves
-          find_link(:href => 'https://reserves.library.nd.edu', :title => 'Course Reserves')
+          find_link(href: 'https://reserves.library.nd.edu', title: 'Course Reserves')
           # ILL
-          find_link(:href => 'https://nd.illiad.oclc.org/illiad/IND/illiad.dll', :title => 'Interlibrary Loan and Document Delivery')
+          find_link(href: 'https://nd.illiad.oclc.org/illiad/IND/illiad.dll', title: 'Interlibrary Loan and Document Delivery')
           # Reserve Room
-          find_link(:href => 'http://nd.libcal.com/#s-lc-box-2749-container-tab1', :title => 'Reserve a Room')
+          find_link(href: 'http://nd.libcal.com/#s-lc-box-2749-container-tab1', title: 'Reserve a Room')
         end
       end
 
@@ -60,8 +59,7 @@ module WebRenovation
       def has_hours?
         # Hours display mid page
         within('.hours-display') do
-          find_link(:href => "/page/hours/").text != ''
-          find_link(:href => "/page/hours/").text != nil
+          !find_link(href: "/page/hours/").text.nil?
         end
       end
     end
--- a/spec/webRenovation/pages/home_page.rb
+++ b/spec/webRenovation/pages/home_page.rb
@@ -8,9 +8,8 @@ module WebRenovation
 
       def on_page?
         super &&
-          on_valid_url? &&
           has_nd_branding? &&
-          searchbox_valid? &&
+          has_valid_searchbox? &&
           has_service_boxes? &&
           has_news? &&
           has_events? &&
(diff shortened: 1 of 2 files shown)


spec/webRenovation/pages/home_page.rb
```diff
@@ -7,17 +7,58 @@ class HomePage < BasePage
       include CapybaraErrorIntel::DSL
 
       def on_page?
-        super && on_valid_url?
+        super &&
+          on_valid_url? &&
+          has_nd_branding? &&
+          searchbox_valid? &&
```
h-parekh: Let's keep method names consistent with others, suggest renaming to something like: 'has_valid_searchbox?'

spec/webRenovation/pages/home_page.rb
```diff
@@ -7,17 +7,58 @@ class HomePage < BasePage
...
 
-      def on_valid_url?
-        current_url == Capybara.app_host || current_url == (Capybara.app_host + "/")
+      def searchbox_valid?
+        find_by_id('searchAppliance') &&
+          find_button('Search')
       end
 
+      def has_service_boxes?
+        # Reserves
```
h-parekh: I'd recommend putting these assertions in a 'within' block. So it doesn't assert on duplicate URLs on the rest of the page. 

``` ruby
within('#reactRoot') do
find_link()
find_link()
find_link()
end 
```

spec/webRenovation/pages/home_page.rb
```diff
@@ -7,17 +7,58 @@ class HomePage < BasePage
       include CapybaraErrorIntel::DSL
 
       def on_page?
-        super && on_valid_url?
+        super &&
+          on_valid_url? &&
+          has_nd_branding? &&
```
h-parekh: This is already being asserted in the super call to ['valid_page_content'](https://github.com/ndlib/QA_tests/blob/WR-301/spec/webRenovation/pages/base_page.rb#L23) method of class BasePage.

spec/webRenovation/pages/home_page.rb
```diff
@@ -7,17 +7,58 @@ class HomePage < BasePage
...
+      def has_service_boxes?
+        # Reserves
+        find_link(:href => 'https://reserves.library.nd.edu', :title => 'Course Reserves')
+        # ILL
+        find_link(:href => 'https://nd.illiad.oclc.org/illiad/IND/illiad.dll', :title => 'Interlibrary Loan and Document Delivery')
+        # Reserve Room
+        find_link(:href => 'http://nd.libcal.com/#s-lc-box-2749-container-tab1', :title => 'Reserve a Room')
+      end
+
+      def has_news?
```
h-parekh: I'd recommend putting these assertions in a 'within' block. So it doesn't assert on duplicate URLs on the rest of the page. 

``` ruby
within('.row.news') do
 find()
end

File: spec/webRenovation/pages/home_page.rb [hunk 5]
```diff
@@ -7,17 +7,58 @@ class HomePage < BasePage
...
+        find('h3', text: 'News') &&
+          all('.news-card', between: 1..3)
+      end
+
+      def has_events?
+        find('h3', text: 'Events') &&
+          all('.event-card', between: 1..3)
+      end
+
+      def has_hours?
```
h-parekh: I'd recommend putting these assertions in a 'within' block. So it doesn't assert on duplicate URLs on the rest of the page. 

``` ruby
          # Hours display mid page
        within('.hours-display') do
          all(:css, 'a[href="/page/hours/"]')[0].text != ''
          all(:css, 'a[href="/page/hours/"]')[0].text != nil
        end

        # Hours button in footer
        within('#footer-info') do
          find_link(:href => '/page/hours/', :class =>'hours')
        end
end

File: spec/webRenovation/pages/home_page.rb [hunk 6]
```diff
@@ -7,17 +7,58 @@ class HomePage < BasePage
...
+        find_link(:href => 'https://reserves.library.nd.edu', :title => 'Course Reserves')
+        # ILL
+        find_link(:href => 'https://nd.illiad.oclc.org/illiad/IND/illiad.dll', :title => 'Interlibrary Loan and Document Delivery')
+        # Reserve Room
+        find_link(:href => 'http://nd.libcal.com/#s-lc-box-2749-container-tab1', :title => 'Reserve a Room')
+      end
+
+      def has_news?
+        find('h3', text: 'News') &&
+          all('.news-card', between: 1..3)
```
h-parekh: Will the page always render at least one card and less than 3 cards?

spec/webRenovation/pages/home_page.rb
```diff
@@ -7,17 +7,58 @@ class HomePage < BasePage
...
+      end
+
+      def has_events?
+        find('h3', text: 'Events') &&
+          all('.event-card', between: 1..3)
```
h-parekh: Will the page always render at least one card and less than 3 cards?

spec/webRenovation/pages/home_page.rb
```diff
@@ -7,17 +7,58 @@ class HomePage < BasePage
...
+        # Reserve Room
+        find_link(:href => 'http://nd.libcal.com/#s-lc-box-2749-container-tab1', :title => 'Reserve a Room')
+      end
+
+      def has_news?
+        find('h3', text: 'News') &&
+          all('.news-card', between: 1..3)
+      end
+
+      def has_events?
```
h-parekh: Use a 'within' block:

``` ruby
within('.row.news') do
 find()
end 
```

spec/webRenovation/pages/home_page.rb
```diff
@@ -7,17 +7,58 @@ class HomePage < BasePage
...
+          all('.event-card', between: 1..3)
+      end
+
+      def has_hours?
+        # Hours display mid page
+        all(:css, 'a[href="/page/hours/"]')[0].text != ''
+        all(:css, 'a[href="/page/hours/"]')[0].text != nil
+
+        # Hours button in footer
+        find_link(:href => '/page/hours/', :class =>'hours')
```
h-parekh: Since this is a footer element visible on all pages of the site, I'm thinking if we should put this validation in the [valid_page_content? method of class BasePage](https://github.com/ndlib/QA_tests/blob/WR-301/spec/webRenovation/pages/base_page.rb#L23).

That way we can 'super' header and footer validations on all other pages.
Thoughts?

spec/webRenovation/pages/home_page.rb
```diff
@@ -7,17 +7,58 @@ class HomePage < BasePage
...
-      def on_valid_url?
-        current_url == Capybara.app_host || current_url == (Capybara.app_host + "/")
+      def searchbox_valid?
+        find_by_id('searchAppliance') &&
+          find_button('Search')
       end
 
+      def has_service_boxes?
+        # Reserves
+        find_link(:href => 'https://reserves.library.nd.edu', :title => 'Course Reserves')
```
h-parekh: Will this URL be exactly the same in all environments? Will this assertion break if I run the test against dev/test?

spec/webRenovation/pages/home_page.rb
```diff
@@ -7,17 +7,58 @@ class HomePage < BasePage
...
+        # ILL
+        find_link(:href => 'https://nd.illiad.oclc.org/illiad/IND/illiad.dll', :title => 'Interlibrary Loan and Document Delivery')
```
h-parekh: Will this URL be exactly the same in all environments? Will this assertion break if I run the test against dev/test?

spec/webRenovation/pages/home_page.rb
```diff
@@ -7,17 +7,58 @@ class HomePage < BasePage
...
+        # Reserve Room
+        find_link(:href => 'http://nd.libcal.com/#s-lc-box-2749-container-tab1', :title => 'Reserve a Room')
```
h-parekh: Will this URL be exactly the same in all environments? Will this assertion break if I run the test against dev/test?

spec/webRenovation/pages/home_page.rb
```diff
@@ -7,17 +7,58 @@ class HomePage < BasePage
...
+      end
+
+      def has_events?
+        find('h3', text: 'Events') &&
+          all('.event-card', between: 1..3)
+      end
+
+      def has_hours?
+        # Hours display mid page
+        all(:css, 'a[href="/page/hours/"]')[0].text != ''
```
h-parekh: Why are you preferring all instead of find_link? Can you document that with a comment?

h-parekh:
@danthewolfe  These are some really good tests!

spec/webRenovation/pages/home_page.rb
danthewolfe: For the moment, that is my assumption. For the page to look right, we need to have an upper limit, and should probably at least show one.

PR #105, opened 2017-04-28
Pathfinder Page Type

hbeachey:
- Add the pathfinder page type for testing, and test one of them. 
- Account page needed updating for tests to work
- Remove extraneous username from base page

### DIFF
--- a/spec/webRenovation/pages/base_page.rb
+++ b/spec/webRenovation/pages/base_page.rb
@@ -6,11 +6,20 @@ module WebRenovation
 
       def on_page?
         status_response_ok? &&
-          valid_page_content?
+          valid_page_content? &&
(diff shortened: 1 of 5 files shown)


spec/webRenovation/pages/base_page.rb
```diff
@@ -25,6 +18,13 @@ def valid_page_content?
           find_link('Hesburgh Libraries').visible?
         end
 
+        if page.current_path != "/" && page.current_path != ""
```
jeremyf: Is this "! on_homepage?" Consider extracting a method on BasePage:

```ruby
def on_homepage?
  page.current_path == "/" || page.current_path == ""
end
```

PR #109, opened 2017-05-18
Mh curate dtlp 1008

mhallwac:
Fixes bug in curate functional test scenario "Show an Article" in regards to to show_article.rb
https://jira.library.nd.edu/browse/DLTP-1008

### DIFF
--- a/spec/curate/pages/show_article.rb
+++ b/spec/curate/pages/show_article.rb
@@ -34,8 +34,8 @@ module Curate
         within("article.abstract.descriptive-text") do
           find('p')
         end
-        within("table.table.table-striped.finding_aid.attributes") do
-          first('p')
+        within("div.work-attributes.span9") do
+	         has_css?('p', minimum: 2)
         end
       end
     end
--- a/spec/curate/pages/show_article.rb
+++ b/spec/curate/pages/show_article.rb
@@ -35,8 +35,9 @@ module Curate
         within("article.abstract.descriptive-text") do
           find('p')
         end
-        within("div.work-attributes.span9") do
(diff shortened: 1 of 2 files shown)


spec/curate/pages/show_article.rb
```diff
@@ -26,16 +27,16 @@ def status_response_ok?
...
+          find("a.btn", text: "View Details")
         end
-        #Make sure that the Abstract and Attributes sections have text
+        # Make sure that the Abstract and Attributes sections have text
         within("article.abstract.descriptive-text") do
           find('p')
         end
-        within("table.table.table-striped.finding_aid.attributes") do
-          first('p')
+        within("div.work-attributes.span9") do
```
jeremyf: I'm not certain what is being tested here. I don't like the `.span9` CSS selector. I believe `div.work-attributes` should be adequate.

spec/curate/pages/show_article.rb
```diff
@@ -26,16 +27,16 @@ def status_response_ok?
...
+	         has_css?('p', minimum: 2)
```
jeremyf: What is this trying to test? We have a minimum of two paragraphs? Is that a safe assumption?

mhallwac: It previously only tested for a first paragraph in the table which would be the name of the creator. Since the Abstract description, which is a paragraph, is included in div.work-attributes, I cannot use the first('p') method but a minimum of 2 ensures that there is a first paragraph in the table.

jeremyf: Please add your response as a comment in the code between line 38 and 39.

PR #113, opened 2017-05-24
Integration test validator with 'hours' API

h-parekh:
* Creating a ResponseValidator class to validate integration test results
* Creates folder structure, configs and spec for 'hours' endpoint

### DIFF
--- a/spec/hours/integration/int_hours_spec.rb
+++ b/spec/hours/integration/int_hours_spec.rb
@@ -1,0 +1,12 @@
+#frozen_string_literal: true
+require 'hours/hours_spec_helper'
+
+describe 'Hours API test' do
+  SwaggerHandler.operations(for_file_path: __FILE__).each do |operation|
+    it "calls #{operation.verb} #{operation.path}" do
+      result = public_send(operation.verb, operation.url)
+      expect(result.code.to_s).to eq(operation.responses.keys[0])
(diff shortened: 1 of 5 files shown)


spec/hours/integration/int_hours_spec.rb
```diff
@@ -0,0 +1,12 @@
+#frozen_string_literal: true
+require 'hours/hours_spec_helper'
+
+feature 'Hours API test' do
+  SwaggerHandler.operations(for_file_path: __FILE__).each do |operation|
```
jeremyf: I believe __FILE__ in this case is "spec/hours/integration/int_hours_spec.rb" Is that the intended file to use for_file_path?

spec/spec_support/response_validator.rb
```diff
@@ -0,0 +1,37 @@
...
+    resolved_schema = response_schema.schema.root.definitions
+    # Resolve definition properties
+    schema_properties = resolved_schema.values[0].properties.keys
+
+    # Parse result body as JSON and find the keys
+    result_keys = JSON.parse(@current_result.body).keys
+
+    # Making sure all the properties listed in Swagger definition
+    # are present in the result body
+    schema_properties.each do |expected_value|
```
jeremyf: I this will get what we need.

```ruby
(schema_properties & result_keys).sort == schema_properties.sort
```

PR #115, opened 2017-05-30
Mh refactor sleep

mhallwac:
Creates a the SleepInjector module which adds a 2 second sleep statement to the end of Capybara click_on method.  This will aid in removing sleep statements from test code as the statements are often necessary after using click_on to ensure the action has occurred.  Also, this module is included into the spec_helper so it can be used by all tests.  The func_curate_spec.rb was used to test this module and sleep statements have been removed from it.


jeremyf:
@mhallwac There are other places where sleep is used. Once this is working, look at refactoring those sleep statements.

mhallwac:
Will do!

PR #116, opened 2017-05-30
Mh sleep refactor

mhallwac:
Changes the sleep_injector.rb file to remove the module format as a class format is simpler. Also, gets rid of overriding the click_on  method since it calls .click so by overriding the .click method, the sleep statement is passed to the click_on method.  Uses method wrapping to override the .click and .trigger methods to add sleep statements.


mhallwac:
Will change the sleep statements in tests on another branch once pulled

PR #118, opened 2017-06-02
Mh option

mhallwac:
Creates if statement in the RequestBuilder that avoids public_send requests for the 'options' verb. Also updates the hours integration test to use RequestBuilder.


spec/spec_support/sleep_injector.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+require 'capybara/node/element'
+# This class adds a sleep statement at the end of the click_on class for
+
+class Capybara::Node::Element
```
jeremyf: Do we still need this?

PR #119, opened 2017-06-05
Updates RequestBuilder to avoid "options" UI verb

mhallwac:
As we do not want to send a request for the "options" verb,
send_security_vs_none now has an outer if statement so that a
pubic_send will only be used for operation verbs other than "options"

### DIFF
--- a/spec/spec_support/request_builder.rb
+++ b/spec/spec_support/request_builder.rb
@@ -9,7 +9,7 @@ class RequestBuilder
 
   # Identifies if the API has security or not then sends request accordingly
   def send_security_vs_none
-    if @current_operation.verb.to_s != "options"
+    if meaningful_method_name?
       # Get schema of the response
       schema = @current_operation.responses.values[0].schema.root
       if schema.keys.include?('securityDefinitions')
@@ -27,5 +27,9 @@ class RequestBuilder
         RestClient.public_send(@current_operation.verb, @current_operation.url)
       end
     end
+  end
+
+  def meaningful_method_name?
+    @current_operation.verb.to_s != "options"
   end
 end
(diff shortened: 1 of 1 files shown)


spec/spec_support/request_builder.rb
```diff
@@ -9,21 +9,23 @@ def initialize(operation)
...
-      name = schema_security.name
-      user_home_dir = File.expand_path('~')
-      # Accesses the file to get the jwt token
-      yaml = YAML.load_file(File.join("#{user_home_dir}", 'test_data/QA/jwt_token.yml'))
-      token = yaml.fetch('token')
-      # Uses the name and token variables bypass security
-      RestClient.public_send(@current_operation.verb, @current_operation.url, "#{name}": "#{token}")
-    else
-      RestClient.public_send(@current_operation.verb, @current_operation.url)
+    if @current_operation.verb.to_s != "options"
```
jeremyf: I like this. Extract a method that encodes this logic; The method name should provide context on the purpose of the if statement.

We'd end up with something like:

```ruby
if a_meaningful_method_name?
```

spec/spec_support/request_builder.rb
```diff
@@ -9,21 +9,27 @@ def initialize(operation)
...
+        token = yaml.fetch('token')
+        # Uses the name and token variables bypass security
+        RestClient.public_send(@current_operation.verb, @current_operation.url, "#{name}": "#{token}")
+      else
+        RestClient.public_send(@current_operation.verb, @current_operation.url)
+      end
     end
   end
+
+  def meaningful_method_name?
```
jeremyf: @mhallwac Please rename `meaningful_method_name?` to a method that describes the intention of the method body. I know I suggested `meaningful_method_name?` but I was hoping you would give it context.

PR #120, opened 2017-06-06
Hp input url

h-parekh:
In AWS code build/deploy, new environment URLs are not static and are built on-the-fly. Currently, our QA framework restricts testing against only those endpoints that are listed in spec/appName/appName_config.yml. This pull request allows us to pass a URL as target environment for testing an application/endpoint which will simplify continuous integration of QA testing with AWS.

In this PR I am also refactoring some previous code to simplify module behavior and intent. 

* Refactoring test_runtime.rb to separate modules
* Creates a new module InitializeExample to set Capybara.app_host based on value of ENVIRONMENT

### DIFF
--- a/spec/spec_support/test_runtime.rb
+++ b/spec/spec_support/test_runtime.rb
@@ -26,7 +26,7 @@ module ScreenshotsManager
     Dir.chdir(screenshots_root)
     screenshots_directories = Dir.glob('*').select { |f| File.directory? f } # Returns a list of all screenshots directories in screenshots_root
     screenshots_directories.reverse.each_with_index { |dir, index|
-      if index>=runs
+      if index >= runs
         FileUtils.rm_rf(dir)
       end
     }
@@ -39,6 +39,8 @@ module ScreenshotsManager
 end
 
 module InitializeExample
+  # * Checks if value of ENVIRONMENT is a URL or key
+  # * Sets Capybara.app_host to a URL
   def self.initialize_app_host(example:, config:)
     @example = example
     @config = config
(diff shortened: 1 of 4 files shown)


spec/spec_support/test_runtime.rb
```diff
@@ -22,14 +26,43 @@ def self.get_screenshots_save_path(runs: 5)
...
+end
+
+module InitializeExample
+  # * Checks if value of ENVIRONMENT is a URL or key
+  # * Sets Capybara.app_host to a URL
+  def self.initialize_app_host(example:, config:)
+    @example = example
+    @config = config
+    initialize_example_variables!
+    if valid_url?(@example_variable.environment_under_test)
```
jeremyf: I agree that we should allow a custom URL that we are going to test. Though this will create havoc if we run all of the tests (instead of tests against a single application).

I do not, however, like overloading the `@example_variable.environment_under_test` with either a URL or a named environment (for which we lookup that value). I can imagine future surprises where our named environment concepts (e.g. `prep`, `prod`, etc) have meaning as to what types of tests can be run (in `prod` we will run only read_only tests). I would rather add a new variable to 

I'll have some time this afternoon to refactor this. I'm content with merging this and working on a refactor. Or adding the refactor to this change as another commit. Or having you do some refactoring. Recommendations?

jeremyf: I suppose my larger question is: would you like me to work with you to refactor these changes? Or would you prefer to review my refactor after I've made it?

h-parekh: Let's merge this, we can sit together this afternoon and I'll watch and learn while you refactor. That will be the easiest way for me to rearrange my approach based on your changes and it will be fast as well.

PR #122, opened 2017-06-08
Mh osf func

mhallwac:
Creates the necessary files and folders for osf tests.  Also updates inject_slep.rb to include the fill_in method as some login fields need time after filling in before clicking submit.  

Note: osf_spec_helper has code forcing the use of tlsv1.2 driver since the OSF website rejects ssl requests.

### DIFF
--- a/spec/osf/functional/func_osf_spec.rb
+++ b/spec/osf/functional/func_osf_spec.rb
@@ -6,5 +6,6 @@ feature 'Institutional Login', js: true do
   scenario 'Sign in by institution (Notre Dame)' do
     login_page.completeLogin
     expect(page).to have_selector('.dropdown-toggle.nav-user-dropdown')
+    # This selector is the user dropdown menu that only appears with a successful login
   end
 end
--- a/spec/osf/functional/func_osf_spec.rb
+++ b/spec/osf/functional/func_osf_spec.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+require 'osf/osf_spec_helper'
+
+feature 'Institutional Login', js: true do
+  let(:login_page) { OSF::Pages::LoginPage.new(current_logger, account_details_updated: false) }
+  scenario 'Sign in by institution (Notre Dame)' do
+    login_page.completeLogin
+    expect(page).to have_link('.dropdown-toggle.nav-user-dropdown')
+  end
+end
--- a/spec/osf/functional/func_osf_spec.rb
+++ b/spec/osf/functional/func_osf_spec.rb
@@ -5,6 +5,6 @@ feature 'Institutional Login', js: true do
   let(:login_page) { OSF::Pages::LoginPage.new(current_logger, account_details_updated: false) }
   scenario 'Sign in by institution (Notre Dame)' do
     login_page.completeLogin
-    expect(page).to have_link('.dropdown-toggle.nav-user-dropdown')
+    expect(page).to have_selector('.dropdown-toggle.nav-user-dropdown')
   end
 end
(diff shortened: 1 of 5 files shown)


spec/osf/functional/func_osf_spec.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+require 'osf/osf_spec_helper'
+
+feature 'Institutional Login', js: true do
+  let(:login_page) { OSF::Pages::LoginPage.new(current_logger, account_details_updated: false) }
+  scenario 'Sign in by institution (Notre Dame)' do
+    login_page.completeLogin
+    expect(page).to have_selector('.dropdown-toggle.nav-user-dropdown')
```
jeremyf: Why does it have this selector? Add a comment to explain your reasoning.

spec/osf/osf_config.yml
```diff
@@ -0,0 +1,5 @@
+local: jello
+pprd: jello
+prod: https://osf.io
+# prod: https://curate.nd.edu/
```
jeremyf: Remove commented out hash key values on line 4 and 5.

spec/osf/pages/login_page.rb
```diff
@@ -0,0 +1,75 @@
...
+
+      def inspect
+        "#<#{self.class} @userName=#{userName.inspect}>"
+      end
+
+      def initialize(logger, account_details_updated: false)
+        @account_details_updated = account_details_updated
+        @current_logger = logger
+        credentials = CSV.read(ENV['HOME']+"/test_data/QA/TestCredentials.csv")
+        credentials = credentials[1, credentials.length-1]
```
jeremyf: Instead of `credentials = credentials[1, credentials.length-1]` you can do `credentials.shift`; Also add a comment about why you are doing this shift.

Also, I believe with CSV you can pass the following:

```ruby
credentials = CSV.read(ENV['HOME']+"/test_data/QA/TestCredentials.csv", headers: true)
```

This will treat the first row as the header row. In using the `header: true` flag you can access elements by their header. Consider the following CSV:

```csv
user,password,hello
jeremy,super,hi
```

```ruby
rows = CSV.read(<The Given CSV>, headers: true)
row = rows.first
row['user'] == 'jeremy'
```

In using the header name, you no longer need worry about the order in which those the columns appear in the csv.

See https://ruby-doc.org/stdlib-2.3.0/libdoc/csv/rdoc/CSV.html for further details.

spec/spec_support/inject_sleep.rb
```diff
@@ -42,3 +42,15 @@ def has_selector?(*args, &optional_filter_block)
...
+
+module Capybara::Node::Actions
+  def fill_in(locator, options={})
+    locator, options = nil, locator if locator.is_a? Hash
+    raise "Must pass a hash containing 'with'" if not options.is_a?(Hash) or not options.has_key?(:with)
+    with = options.delete(:with)
+    fill_options = options.delete(:fill_options)
+    options[:with] = options.delete(:currently_with) if options.has_key?(:currently_with)
+    find(:fillable_field, locator, options).set(with, fill_options)
+    sleep(2)
```
jeremyf: Why does this sleep exist?

spec/osf/pages/login_page.rb
mhallwac: I did not use the header: true option because then it is not an array and we want to use the sample method for arrays to randomize which combination of test login info is used.

spec/spec_support/inject_sleep.rb
mhallwac: This sleep is here because when clicking a button after a fill_in statement, the click statement does not always click properly. This gives the fill_in statement time to completely finish and solves this click issue.

jeremyf: @mhallwac Please copy your response and add it as a comment above the sleep method call.

PR #125, opened 2017-06-12
Mh login page

mhallwac:
Creates a global cas_login.rb page for tests to use in logged in scenarios. As many tests have a login page in the folder, creating a global login page allows for consistency across different tests and reduces amount of code needed.  This page also keeps track of if the a value was passed by the test for account_details_updated to use only the proper test credentials. The functional test for curate has ben updated to utilize this global login page and the curate/pages/login_page.rb has been updateed to only add the checkLoginPage method for the LoginPage class for the curate folder. Inject_sleep.rb has also been updated to include the has_no_selector? method because a sleep statement is sometimes necessary when testing if the completeLogin method logs in and loads the logged in homepage.

### DIFF
--- a/spec/curate/functional/func_curate_spec.rb
+++ b/spec/curate/functional/func_curate_spec.rb
@@ -158,7 +158,6 @@ feature 'Logged In User (Account details NOT updated) Browsing', js: true do
     visit '/'
     click_on('Log In')
     login_page.completeLogin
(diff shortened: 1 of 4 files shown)


jeremyf:
@mhallwac Looks like you have a merge conflict to resolve.

mhallwac:
Just fixed the merge conflict

PR #126, opened 2017-06-14
Mh verify network

mhallwac:
Updates test_runtime.rb to skip verify_network_traffic for integration tests as it is unnecessary for them. Also adds needed 'end' statements to inject_sleep.rb to make it functional.  Lastly, fixes the call to ExampleVariableExtractor in swagger_handler.rb since config: is necessary when calling that class.

### DIFF
--- a/spec/spec_support/test_runtime.rb
+++ b/spec/spec_support/test_runtime.rb
@@ -105,9 +105,16 @@ module VerifyNetworkTraffic
     @test_handler = test_handler
     return true unless driver_allows_network_traffic_verification?
     return true if ENV.fetch('SKIP_VERIFY_NETWORK_TRAFFIC', false)
+    return true if test_needs_network_verification?
     ExampleLogging.current_logger.info(context: "verifying_all_network_traffic") do
       verify_network_traffic(driver: driver)
     end
+  end
+
+  def self.test_needs_network_verification?
+    input = ARGV.join.split('/')
+    #splits the test path into individual array elements
+    input.include?('integration')
   end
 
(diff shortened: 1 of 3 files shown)


spec/spec_support/test_runtime.rb
```diff
@@ -105,16 +105,24 @@ def self.report_network_traffic(driver:, test_handler:)
     @test_handler = test_handler
     return true unless driver_allows_network_traffic_verification?
     return true if ENV.fetch('SKIP_VERIFY_NETWORK_TRAFFIC', false)
+    return true if test_needs_network_verification?
```
jeremyf: In reading this, I think the name of `test_needs_network_verification?` is incorrect. I believe it should be ` return true unless test_needs_network_verification?`

spec/spec_support/test_runtime.rb
```diff
@@ -105,16 +105,24 @@ def self.report_network_traffic(driver:, test_handler:)
...
+    #splits the test path into individual array elements
+    input.include?('integration')
+  end
+
   def self.driver_allows_network_traffic_verification?
     Capybara.current_driver == :poltergeist
   end
 
   def self.verify_network_traffic(driver:)
+    require 'byebug'; debugger
```
jeremyf: Remove please!

spec/spec_support/test_runtime.rb
mhallwac: I reversed the logic in the method and changed it to unless, the statement should make more sense now

PR #127, opened 2017-06-14
Changes to crawler based on ESU requirements

h-parekh:
These changes are adding some robustness to the site crawler based on requirements from ESU team. 
* Additional logging: being able to see which URLs are being hit.
* Showing a prompt to create a local input file
* Catching timeout to unreachable links and continue with rest of spec

I realize this isn't the best implementation based on our current ecosystem. This test works for now and we can refactor at a later point.

### DIFF
--- a/spec/crawler/crawl/web_crawler_spec.rb
+++ b/spec/crawler/crawl/web_crawler_spec.rb
@@ -6,44 +6,49 @@ require "uri"
 require "set"
 feature 'Link Checker' do
   scenario 'Check Links' do
-    File.foreach(ENV["HOME"] + "/crawl_sites.txt") do |root_url|
-      root_url.strip!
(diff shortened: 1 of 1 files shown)


spec/crawler/crawl/web_crawler_spec.rb
```diff
@@ -6,29 +6,49 @@
...
-      for href in hrefs do
-        if href.start_with?('/')
-          href = root_url + href
+    begin
+      File.foreach(ENV["HOME"] + "/crawl_sites.txt") do |root_url|
+        root_url.strip!
+        current_logger.info(context: "Begin verifying site: #{root_url}")
+        encoded_url = URI.encode(root_url)
+        doc = Nokogiri::HTML(open(encoded_url))
+        links = doc.css('a')
```
dbrower: Maybe also include js, css, and other files referenced in the head of the page?

PR #132, opened 2017-06-15
Splits request builder into different methods

mhallwac:
RequestBuilder is now split into different methods to clarify the code.
In addition, a method_case method was added to create a case-when series
based on the operation.verb request (e.g. get, post). This method has an
optional argument *body as certain operation.verbs (such as post, put
) need a body to added in the public_send request.  RequestBuilder
also now has a logger as an instance variable so the status of the
request can be logged.  The hours and gatekeeper int. tests have been
updated to use the new RequestBuilder.

### DIFF
--- a/spec/spec_support/request_builder.rb
+++ b/spec/spec_support/request_builder.rb
@@ -5,7 +5,7 @@ require 'csv'
 
 class RequestBuilder
   attr_reader :current_logger
-  
+
   def initialize(logger,operation)
     @current_operation = operation
     @security_name = get_security_name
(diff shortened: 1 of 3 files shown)


spec/spec_support/request_builder.rb
```diff
@@ -1,35 +1,82 @@
...
+  def get_security_name
+    # Get schema of the response
+    schema = @current_operation.responses.values[0].schema.root
+    # Get the type, name, and in fields of the security
+    schema_security = schema.securityDefinitions.values[0]
+    # Gets the name of the security to use for authorization
+    name = schema_security.name
+  end
+
+  def method_case(*body)
```
jeremyf: Can you explain `method_case` ? Is it performing the REST request? The method name is somewhat opaque and not something I quickly understand when I read the following:

```ruby
schema = RequestBuilder.new(current_logger, operation)
result = schema.method_case
```

PR #138, opened 2017-06-22
Hp usurper tests

mhallwac:
Integration test for the usurper content api.  Creates, previews, then deletes an entry in contentful in  order to test the api. Uses https://alpha.library.nd.edu as its prod config because it visits the website with the name of the entry as a slug to make sure the entry is previewed.

### DIFF
--- a/spec/spec_support/contentful_handler.rb
+++ b/spec/spec_support/contentful_handler.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+# This class provides support to interact with the following contentful APIs:
+# * Content Delivery API: https://www.contentful.com/developers/docs/references/content-delivery-api/
+# * Content Management API: https://www.contentful.com/developers/docs/references/content-management-api/
+# * Content Preview API: https://www.contentful.com/developers/docs/references/content-preview-api/
+# * Images API: https://www.contentful.com/developers/docs/references/images-api/
+
+class ContentfulHandler
+  def initialize(for_file_path:, config:, current_logger:)
+    @current_logger = current_logger
+    example_variable = ExampleVariableExtractor.call(path: for_file_path, config: config)
+    read_contentful_tokens
+    create_entry
+  end
+
+  def read_contentful_tokens
+    user_home_dir = File.expand_path('~')
+    cf_key_file = YAML.load_file(File.join(user_home_dir.to_s, 'test_data/QA/cf_api_key.yml'))
+    @space_id = cf_key_file.fetch('QA_key').fetch('space_id')
+    @cdn_token = cf_key_file.fetch('QA_key').fetch('cdn_token')
+    @preview_token = cf_key_file.fetch('QA_key').fetch('preview_token')
+    @personal_access_token = cf_key_file.fetch('QA_key').fetch('personal_access_token')
+  end
+
+  def create_entry
+    create_client
+    content_type = @client.content_types.find(@space_id, 'page')
+    @title_for_entry = 'Testing page ' + RunIdentifier.get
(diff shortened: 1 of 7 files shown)


spec/spec_support/contentful_handler.rb
```diff
@@ -0,0 +1,61 @@
...
+# * Content Delivery API: https://www.contentful.com/developers/docs/references/content-delivery-api/
+# * Content Management API: https://www.contentful.com/developers/docs/references/content-management-api/
+# * Content Preview API: https://www.contentful.com/developers/docs/references/content-preview-api/
+# * Images API: https://www.contentful.com/developers/docs/references/images-api/
+
+class ContentfulHandler
+  attr_reader :slug_for_entry
+  def initialize(for_file_path:, config:, current_logger:)
+    @current_logger = current_logger
+    example_variable = ExampleVariableExtractor.call(path: for_file_path, config: config)
```
jeremyf: Why is this being set?

spec/spec_support/contentful_handler.rb
```diff
@@ -0,0 +1,61 @@
...
+    read_contentful_tokens
+    create_entry
+  end
+
+  def read_contentful_tokens
+    user_home_dir = File.expand_path('~')
```
jeremyf: Instead of "~" which may not work on Windows, prefer `ENV['HOME']`

```ruby
cf_key_file = YAML.load_file(File.join(ENV.fetch('HOME'), 'test_data/QA/cf_api_key.yml'))
```

spec/spec_support/contentful_handler.rb
```diff
@@ -0,0 +1,61 @@
...
+  def create_entry
+    create_client
+    @content_type = @client.content_types.find(@space_id, 'page')
+    @title_for_entry = 'Testing_page_' + RunIdentifier.get
+    @slug_for_entry = @title_for_entry
+    @current_logger.info(context: "Creating page in contentful", page_title: @title_for_entry)
+    @entry = @client.entries.create(@content_type, title: @title_for_entry, slug: @slug_for_entry)
+  end
+
+  def create_client
```
jeremyf: Make this a private method.

spec/usurper/integration/int_usurper_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+
+feature 'Test for Usurper content management API' do
+  scenario "Creates, previews, and deletes an entry" do
+    created_entry = ContentfulHandler.new(for_file_path: __FILE__, config: ENV, current_logger: current_logger)
+    expect(created_entry).to be_in_contentful
+    # Write a function to preview the entry just created using Content Preview API
+    created_entry.make_entry_previewable
+    visit "/#{created_entry.slug_for_entry}?preview=true"
+    # Remove the entry created using the Content Management API
+    created_entry.delete_entry
```
jeremyf: I'm thinking that there is a potential problem. If anything fails after created_entry is made, but before it is deleted, we could end with an orphaned item.

When you are ready to review this, let's work on it together. We'll likely need to use something along the lines of `ensure`. For reference look at The first entry on https://stackoverflow.com/questions/4224587/file-open-with-block-vs-without (that starts with "DarkDust already said that these methods are different.")

spec/usurper/usurper_spec_helper.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+require 'spec_helper'
+require 'contentful/management'
+require 'contentful'
```
jeremyf: As a matter of preference, place specific requires after general requires. In other words

```ruby
require 'spec_helper'
require 'contentful'
require 'contentful/management'
```

PR #139, opened 2017-06-22
Adding scripts to generate audit reports

jeremyf:


bin/run_lighthouse_against_urls_and_generate_csv.rb
```diff
@@ -0,0 +1,23 @@
+# Generate a series of JSON files based on Lighthouse audits.
+#
+# @see https://github.com/GoogleChrome/lighthouse for instructions on installing the lighthouse script
```
h-parekh: Probably need these instructions in README.me as well,

bin/consolidate_lighthouse_json_to_csv.rb
```diff
@@ -0,0 +1,48 @@
...
+DATE_SUFFIX = Time.now.strftime('--on-%Y-%m-%d-at-%H-%M')
+CSV.open(PROJECT_PATH.join("tmp/accessibility#{DATE_SUFFIX}.csv"), 'w+') do |csv|
+  csv << ["URL", "CATEGORY_NAME", "AUDIT_ID", "AUDIT_HELP", "NOTE", "SELECTOR"]
+  Dir.glob(INPUT_PATH.join("**/*.json")).each do |filename|
+    json = JSON.parse(File.read(filename))
+    url = json.fetch('initialUrl')
+    json.fetch('reportCategories').each do |report_category|
+      category_name = report_category.fetch('name')
+      next unless REPORT_CATEGORIES_TO_TEST.include?(category_name)
+      next if report_category.fetch("score") >= 100
```
h-parekh: It looks like you're skipping those items which are not violating lighthouse rules? 
Since this is an audit it would be good to have the list of items that passed as well. So we can keep a track over multiple runs when/if they fail.

bin/consolidate_lighthouse_json_to_csv.rb
```diff
@@ -0,0 +1,48 @@
...
+      report_category.fetch("audits").each do |audit|
+        next if audit.fetch('score') >= 100
```
h-parekh: Same comment, can we save the items that pass as well?

bin/consolidate_lighthouse_json_to_csv.rb
jeremyf: The raw JSON is captured and stored. There won't be any meaningful items within the document; This is a short-circuit. I'll add a comment to explain the short-circuit.

PR #144, opened 2017-06-26
Mh database

mhallwac:
Creates the database check to generate a list of missing databases from the website.  Uses the new gem google_drive to access google sheets and compare the list of needed databases from google sheets with the list of databases on the website.


spec/webRenovation/lib/database_check_spec.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+require 'webRenovation/webRenovation_spec_helper'
+require 'google_drive'
+
+session = GoogleDrive::Session.from_config("config.json")
```
jeremyf: This should be in a `let` block (otherwise it will always run). And it should be moved to the line just after `feature 'User Browsing'`

```ruby
let(:session) { GoogleDrive::Session.from_config("config.json") }
let(:ws_hash) do
  ws = session.spreadsheet_by_key("1sUEizFQYu1fc0kunEST-JpHWWH9vMWHr0lj1ZKiIWyk").worksheets[1]
  ws_hash = {}
  for i in 2..ws.num_rows do
    ws_hash[ws[i,2]] = ws[i,1]
  end
  ws_hash
end
```

config.json
```diff
@@ -0,0 +1,9 @@
+{
+  "client_id": "452925651630-egr1f18o96acjjvphpbbd1qlsevkho1d.apps.googleusercontent.com",
+  "client_secret": "1U3-Krii5x1oLPrwD5zgn-ry",
```
jeremyf: I don't like secrets from this page…it should not be committed in the repository. Nor do I like the `refresh_token`

PR #146, opened 2017-06-26
ResponsiveHelpers module for changing window size

h-parekh:
This module provides methods to set the window size for devices in
landscape and portrait modes.
By default, I'm setting the window size to 1280x800 in landscape
which simulates desktop mode.

### DIFF
--- a/spec/spec_support/responsive_helper.rb
+++ b/spec/spec_support/responsive_helper.rb
@@ -9,11 +9,9 @@ module ResponsiveHelpers
   end
 
   def self.resize_window_default(mode)
-    require 'byebug'; debugger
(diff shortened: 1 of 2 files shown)


spec/spec_support/responsive_helper.rb
```diff
@@ -0,0 +1,24 @@
...
+  def self.resize_window_to_mobile(mode)
+    resize_window_by([640, 480], mode)
+  end
+
+  def self.resize_window_to_tablet(mode)
+    resize_window_by([1024, 768], mode)
+  end
+
+  def self.resize_window_default(mode)
+    require 'byebug'; debugger
```
h-parekh: Sorry about that!

PR #153, opened 2017-06-28
Hp gatekeeper test

h-parekh:
* Checks for a path parameter and applies it to the path
* Adds additional logging 
* Fixes Rubocop warnings

### DIFF
--- a/spec/spec_support/request_builder.rb
+++ b/spec/spec_support/request_builder.rb
@@ -11,7 +11,7 @@ class RequestBuilder
     @security_name = get_security_name
     @token = get_token
     @current_logger = logger
-    check_for_path_parameter!
+    apply_parameter_to_parameterized_path!
   end
 
   # Identifies if the API has security or not then sends request accordingly
@@ -20,7 +20,7 @@ class RequestBuilder
     schema.keys.include?('securityDefinitions')
   end
 
-  def check_for_path_parameter!
+  def apply_parameter_to_parameterized_path!
     path_parameter = @current_operation.path.match(/\{(.*)\}/)
     if path_parameter
       current_logger.debug(context: "Operation contains path parameter", path: @current_operation.path)
--- a/spec/spec_support/request_builder.rb
+++ b/spec/spec_support/request_builder.rb
@@ -6,7 +6,7 @@ require 'csv'
 class RequestBuilder
   attr_reader :current_logger
 
-  def initialize(logger,operation)
+  def initialize(logger, operation)
     @current_operation = operation
     @security_name = get_security_name
     @token = get_token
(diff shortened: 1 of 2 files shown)


spec/spec_support/request_builder.rb
```diff
@@ -6,11 +6,12 @@
 class RequestBuilder
   attr_reader :current_logger
 
-  def initialize(logger,operation)
+  def initialize(logger, operation)
     @current_operation = operation
     @security_name = get_security_name
     @token = get_token
     @current_logger = logger
+    check_for_path_parameter!
```
jeremyf: Not a fan of the name `check_for_path_parameter!` method name. I prefer `apply_parameter_to_parameterized_path!`

spec/spec_support/request_builder.rb
```diff
@@ -19,11 +20,29 @@ def check_security
     schema.keys.include?('securityDefinitions')
   end
 
+  def check_for_path_parameter!
+    path_parameter = @current_operation.path.match(/\{(.*)\}/)
+    if path_parameter
+      current_logger.debug(context: "Operation contains path parameter", path: @current_operation.path)
+      sampled_parameter = get_path_parameter!
+      @current_operation.path = @current_operation.path.sub(path_parameter[0], sampled_parameter[0])
```
jeremyf: Add a comment saying that we assume there is only one parameter in the path.

spec/spec_support/request_builder.rb
```diff
@@ -19,11 +20,29 @@ def check_security
...
+      current_logger.debug(context: "Path parameter has been updated", path: @current_operation.path)
+    else
+      current_logger.debug(context: "Operation does not contain a path parameter", path: @current_operation.path)
+    end
+  end
+
+  def get_path_parameter!
+    user_home_dir = File.expand_path('~')
+    parameter_list = CSV.read(user_home_dir + "/test_data/QA/aleph_system_ids.csv")
```
jeremyf: How do we know that we should use this CSV? There is some degree of hard-coding that could be a serious surprise (I assume the operations are used across different projects).

Should we create a YAML file that has the parametrized path as a key, then have an array of replacements?

spec/spec_support/request_builder.rb
```diff
@@ -33,44 +52,44 @@ def get_security_name
...
+      current_logger.info(context: "making POST request", url: @current_operation.url)
       if security
         RestClient.public_send(@current_operation.verb, @current_operation.url, body, "#{@security_name}": "#{@token}")
       else
         RestClient.public_send(@current_operation.verb, @current_operation.url, body)
       end
-    when "put" #might need  specific item when testing
-      current_logger.info(context: "naking PUT request")
+    when "put" # might need  specific item when testing
+      current_logger.info(context: "naking PUT request", url: @current_operation.url)
```
jeremyf: Update "naking" with "making"

spec/spec_support/request_builder.rb
h-parekh: Agree, this code should not live in RequestBuilder. I'm thinking of having the request builder  reach out to application's spec for the value to be substituted.
Then we can leave it up to the application's spec to supply the parameter to be substituted.

spec/spec_support/request_builder.rb
h-parekh: Good catch! Will want to accommodate logic for multiple path parameters.

jeremyf:
Further work captured here #156

PR #158, opened 2017-06-29
Mh webren

mhallwac:
* new scenario tests the functionality of the default OneSearch on the
home page
* new SearchPage class tests if the OneSearch page loads fully, gets
200 status, and if the url is right.
* sleep statement is necessary due to the time it takes for the page to load
* creates the new scenario for using one of the non-default search
methods (ND Catalog)
* tests to see if the search page fully loads with the correct url and status
code
* updates the SearchPage class to accommodate non-OneSearch search methods

### DIFF
--- a/spec/webRenovation/functional/func_webRenovation_spec.rb
+++ b/spec/webRenovation/functional/func_webRenovation_spec.rb
@@ -84,10 +84,24 @@ feature 'User Browsing', js: true do
     page.driver.browser.js_errors = false
     visit '/'
     find('#header-search-button').click
-    expect(page).to have_selector('.current-search')
     find_button('Search').trigger('click')
     search = WebRenovation::Pages::SearchPage.new
-    # waits for search to load fully 
+    # waits for search to load fully
+    sleep(1)
+    expect(search).to be_on_page
+  end
+
+  scenario 'Search using ND Catalog from HomePage' do
+    page.driver.browser.js_errors = false
+    visit '/'
+    find('#header-search-button').click
+    find('.current-search').click
+    within('.uSearchOptionList') do
+      find('p', text: 'ND Catalog').click
+    end
+    find_button('Search').trigger('click')
+    search = WebRenovation::Pages::SearchPage.new
+    # waits for search to load fully
     sleep(1)
     expect(search).to be_on_page
   end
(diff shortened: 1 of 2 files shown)


spec/webRenovation/functional/func_webRenovation_spec.rb
```diff
@@ -79,6 +79,32 @@
     calendar = WebRenovation::Pages::CalendarPage.new
     expect(calendar).to be_on_page
   end
+
+  scenario 'Search using OneSearch from HomePage' do
+    page.driver.browser.js_errors = false
```
jeremyf: Not a blocker, but why are you disabling JS errors? I'd personally like them to be on.

spec/webRenovation/pages/search_page.rb
```diff
@@ -0,0 +1,42 @@
...
+        page.has_selector?('#exlidUserAreaRibbon') &&
+        page.has_selector?('#exlidMainMenuRibbon') &&
+        page.has_selector?('.tab.onesearch.EXLSearchTabTitle.EXLSearchTabLABELOneSearch') &&
+        page.has_selector?('.tab.ndcatalog.EXLSearchTabTitle.EXLSearchTabLABELND.Campus') &&
+        page.has_selector?('#tab_art_l.tab') &&
+        page.has_selector?('#tab_jou_l.tab') &&
+        page.has_selector?('#tab_dat_l.tab') &&
+        page.has_selector?('.tab.EXLSearchTabTitle.EXLSearchTabLABELeBooks') &&
+        page.has_selector?('#goButton') &&
+        if current_url == 'http://onesearch.library.nd.edu/primo_library/libweb/action/dlSearch.do?bulkSize=10&dym=true&highlight=true&indx=1&institution=NDU&mode=Basic&onCampus=false&pcAvailabiltyMode=true&query=any%2Ccontains%2C&search_scope=nd_campus&tab=nd_campus&vid=NDU&displayField=title&displayField=creator'
```
jeremyf: This URL scares me. Please make a descriptive constant that explains what this URL. (e.g. This is the principle of no magic strings; name strings that have conditional logic).

spec/webRenovation/functional/func_webRenovation_spec.rb
h-parekh: I agree, I'll remove these on all our tests.

spec/webRenovation/pages/search_page.rb
mhallwac: I gave them variable names of one_search_url and nd_catalog_search_url

PR #167, opened 2017-07-05
Mh web

mhallwac:
These scenarios test the functionality of the CurateND and Library
Website search options by checking to see if the user is brought to
the correct url after searching.

### DIFF
--- a/spec/spec_support/inject_sleep.rb
+++ b/spec/spec_support/inject_sleep.rb
@@ -79,6 +79,19 @@ end
 class RSpec::Matchers::BuiltIn::Eq
   def match(expected, actual)
     actual == expected
+  rescue Capybara::ExpectationNotMet, Exception => e
+    # if error
     sleep(3)
+    begin
+      actual == expected
+    rescue Capybara::ExpectationNotMet
+      # if still error
+      sleep(5)
+      begin
+        actual == expected
+      rescue Capybara::ExpectationNotMet
+        return false
+      end
+    end
   end
 end
--- a/spec/spec_support/inject_sleep.rb
+++ b/spec/spec_support/inject_sleep.rb
@@ -75,3 +75,10 @@ module Capybara::Node::Actions
     # solves the issue of nonworking click statements after fill_in statements
   end
 end
+
+class RSpec::Matchers::BuiltIn::Eq
+  def match(expected, actual)
+    actual == expected
+    sleep(3)
+  end
+end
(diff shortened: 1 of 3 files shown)


spec/webRenovation/functional/func_webRenovation_spec.rb
```diff
@@ -98,16 +96,24 @@
...
+  scenario 'Search using Library Website from HomePage' do
+    visit '/'
+    find('.current-search').click
+    within('.uSearchOptionList') do
+      find('p', text: 'Library Website').click
     end
     find_button('Search').trigger('click')
-    search = WebRenovation::Pages::SearchPage.new
-    # waits for search to load fully
     sleep(1)
```
h-parekh: Consider using sleep injector instead of sleep statements.

spec/webRenovation/functional/func_webRenovation_spec.rb
```diff
@@ -98,16 +96,24 @@
...
   scenario 'Search using CurateND from HomePage' do
     visit '/'
-    find('#header-search-button').click
     find('.current-search').click
     within('.uSearchOptionList') do
-      find('p', text: 'ND Catalog').click
+      find('p', text: 'CurateND').click
+    end
+    find_button('Search').trigger('click')
+    sleep(2)
```
h-parekh: Consider using sleep injector instead of sleep statements.

spec/spec_support/inject_sleep.rb
```diff
@@ -75,3 +75,10 @@ def fill_in(locator, options={})
     # solves the issue of nonworking click statements after fill_in statements
   end
 end
+
+class RSpec::Matchers::BuiltIn::Eq
+  def match(expected, actual)
+    actual == expected
+    sleep(3)
```
h-parekh: This implementation will make every call of eq sleep for 3 seconds. This has 2 drawbacks:
* If response is back in less than 3 seconds, the code makes you wait for a full 3 seconds. This is a lot of overhead considering how extensively we call the 'eq' method across all tests.
* If response takes more than 3 seconds, the implemetnation fails. this defeats the purpose of having a sleep injector. 

Consider an implementation similar to has_selector?
```ruby
module RSpec::Matchers::BuiltIn::Eq
def match(expected, actual)
    actual == expected
  rescue Capybara::ExpectationNotMet, Exception => e
    # if error
    sleep(3)
    begin
      actual == expected
    rescue Capybara::ExpectationNotMet
      # if still error
      sleep(10)
      begin
       actual == expected
      rescue Capybara::ExpectationNotMet
        return false
      end
    end
  end
end
```

spec/webRenovation/pages/search_page.rb
```diff
@@ -10,6 +10,7 @@ def initialize
       end
 
       def on_page?
+        sleep(2)
```
h-parekh: Let's not use sleep statements anywhere outside sleep injector.

PR #194, opened 2017-07-20
Alters the response_validator to check for levels of $refs

mhallwac:
As some APIs have other $refs in the definition of another $ref, the
ResponseValidator now checks for those deeper levels of $refs to get keys for
all of the definitions so that the API result is fully tested against
all of the expected keys. The test does the same thing for the results to make
sure all of the keys are accounted for.

The config file also no has the proper location for the API


h-parekh:
@jeremyf Could you help review this code change please??

PR #209, opened 2017-08-08
Enhancement for SwaggerHandler to read AWS generated swagger defs 

h-parekh:
We're trying to use the same swagger definition for all 'DevOps' activities. ([Here's](https://github.com/ndlib/usurper_content/blob/master/definitions/swagger.yml) a sample)
This enhancement makes the SwaggerHandler run a deep search into a yaml file generated by AWS. 

After I finished up this commit, I came across a [DeepFetch method](https://github.com/intridea/hashie#deepfetch) from Hashie gem which could perhaps do something similar. Any thoughts if I should pursue this as a better solution?


jeremyf:
Hashie's deep fetch would be a good thing. However, let's get this merged.

PR #222, opened 2017-09-15
Updates cas_login.rb to (semi)fix login bug

mhallwac:
**This is a temporary fix**
For some reason, when logging into CAS, the trigger('click') on the login
button of the final step is not firing.  This commit alters this final step
by using a do-until loop that clicks the login button until the CAS login
button (and therefore the page) is no longer visible. Thus, this also utilizes the maleficent
sleep injector to wait in between each click as that may be part of the issue. In doing so, it ensures that the login button will be clicked and keep this bug from causing tests to fail.


jeremyf:
@mhallwac Excellent commit message. Gave me clarity into why the change was made.

PR #226, opened 2017-09-29
The version of bunyan capybara used to use the path parameter

mhallwac:
The Bunyan module now has an path: parameter for its instantiate_all_loggers
method. This parameter is used in the spec_helper to give the location of the
spec folder to the module so it can properly create/update the bunyan_logs folder
and the *.log files. the spec_path variable is used to hold the path structure
to the said spec folder.


mhallwac:
https://github.com/ndlib/bunyan_capybara/pull/2

If you want to see the update to the bunyan_capybara gem

PR #228, opened 2017-10-09
Adding new scenario for 'ORCID settings'

h-parekh:
This scenario logs in with a test user account whose account details have not been updated and checks if the user can get to the ORCID sign up page. 

I'm not testing further scenarios due to following restrictions:
* The ORCID sign up page has a reCAPTCHA that's making it very hard to register as a bot
* Once an account's ORCID id has been created I cannot recreate another one for it. I have only 5 test accounts that can login into CAS, which will get exhausted in few iterations of testing.

### DIFF
--- a/spec/curate/pages/orcid_home_page.rb
+++ b/spec/curate/pages/orcid_home_page.rb
@@ -8,11 +8,12 @@ module Curate
       def on_page?
         on_valid_url? &&
           status_response_ok? &&
-          valid_page_content?
+          valid_page_content? &&
+          valid_uri_parameters?
       end
 
       def on_valid_url?
-        current_url.include?('orcid.org')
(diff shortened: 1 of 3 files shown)


spec/curate/pages/orcid_home_page.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+module Curate
+  module Pages
+    class OrcidHomePage
+      include Capybara::DSL
+      include CapybaraErrorIntel::DSL
+
+      def on_page?
```
jgondron: Can you add a check for the correct query params? 

"redirect_uri" should be "Capybara.app_host/orcid/create_orcid"
"response_type" should be "code"
"scope" should be "/read-limited+/activities/update"

All of these must be present.

PR #232, opened 2017-10-27
Refactoring hours api integration tests and adding contentful scenario 

h-parekh:
* Refactoring 'hours' spec files to 'monarchLibguides' so they align with the repository name
* Adding a new scenario in contentful tests which creates an entry of type 'event'
* This is a peculiar scenario that needs to wait for the page to get updated from libcal after which the @entry object is refreshed and previewed.

I was originally intending for this to live in spec/monarchLibguides but eventually saw a lot of commonalities to keep the scenario in spec/contentful


spec/contentful/integration/int_contentful_spec.rb
```diff
@@ -1,16 +1,38 @@
...
+      expect(entry).to be_deleted
+    end
+  end
+
+  # This scenario is testing the /newevent endpoint of monarch_libguides API
+  # https://github.com/ndlib/monarch_libguides/blob/master/deploy/gateway.yml
+  # I am calling /newevent via contentful to make sure the webhooks are firing
+  # correctly, and the event is then previewable
+  scenario "Create and preview an entry of type 'Events'" do
+    ContentfulHandler.create(current_logger: current_logger, content_type: 'event', lib_cal_id: '3596276') do |entry|
```
jeremyf: What is the magic string `'3596276'`? Please add a comment on the line between 25 and 26 or create a constant that explains what 3596276 is.

spec/contentful/integration/int_contentful_spec.rb
```diff
@@ -1,16 +1,38 @@
...
+      entry.verify_webhooks
+      expect(entry).not_to be_published
+      sleep(10)
```
jeremyf: Add a comment on why we are sleeping for 10 seconds.

spec/contentful/integration/int_contentful_spec.rb
```diff
@@ -1,16 +1,38 @@
...
+      # At this point the event entry should have been updated in contentful, so we need to refresh it
+      entry.refresh_entry
+      entry.make_previewable!
+      visit "/event/#{entry.slug}?preview=true"
+      event_entry_preview = ContentfulTests::Pages::EventEntryPreview.new(contentful_entry: entry)
```
jeremyf: I thought `new(contentful_entry: entry)` was passing `entry` as a keyword. It looks like it is passing a hash. Which complicates the `EventEntryPreview` initialization (more later).

spec/contentful/pages/event_entry_preview.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+module ContentfulTests
+  module Pages
+    class EventEntryPreview
+      include Capybara::DSL
+      include CapybaraErrorIntel::DSL
+
+      def initialize(contentful_entry)
```
jeremyf: As written, the parameter is un-named. And you are passing a hash on initialization. Which means you then need to fetch the key from the hash. I believe you want the parameter to be a keyword parameter. Switch to this:

```ruby
def initialize(contentful_entry:)
  @contentful_entry = contentful_entry
end
```

Which means you can keep doing the following:

```ruby
ContentfulTests::Pages::EventEntryPreview.new(contentful_entry: entry)
```

spec/spec_support/contentful_handler.rb
```diff
@@ -39,20 +46,22 @@ def initialize(current_logger:)
...
   end
 
-  def create_entry(title:, slug:, content_type:)
+  def create_entry(title:, slug:, content_type:, libCalId:)
     current_logger.info(context: "Finding content type '#{content_type}' in contentful", content_type: content_type)
     contentful_content_type = client.content_types.find(@space_id, content_type)
-    current_logger.info(context: "Creating page in contentful", page_title: title)
-    entry = client.entries.create(contentful_content_type, title: title, slug: slug)
-    current_logger.info(context: "Created page in contentful", page_title: title, contentful_entry_id: entry.id)
+    current_logger.info(context: "Creating entry of type #{content_type} in contentful", page_title: title)
```
jeremyf: I notice that you are not logging `libCalId`. Would this be helpful in your review of the build?

spec/spec_support/contentful_handler.rb
```diff
@@ -39,20 +46,22 @@ def initialize(current_logger:)
...
     ContentfulEntryWrapper.new(entry: entry, handler: self)
   end
 
   private
 
     def read_contentful_tokens
       cf_key_file = YAML.load_file(File.join(ENV.fetch('HOME'), 'test_data/QA/cf_api_key.yml'))
-      qa_key = cf_key_file.fetch('QA_key')
+      # Change the qa_key to fetch 'prep' when you want to test in non-production contentful space
+      qa_key = cf_key_file.fetch('prod')
```
jeremyf: Hmm…this looks really suspicious. How can we parameterize that? I'm concerned that I would change from 'prod' to 'prep' in code. It's an opaque option.

(This is not a blocker, but something to write an issue about and refactor)

spec/spec_support/contentful_handler.rb
```diff
@@ -120,6 +153,9 @@ def verify_webhooks
         if found_beta_webhook == false
           current_logger.error(context: "Webhook missing for beta site")
         end
+        if found_libguides_webhook == false
```
jeremyf: Is it always an error (according to logging) when `found_libguides_webhook == false`? I'm not aware of when all we'd be setting (or not setting the webhook).

spec/spec_support/contentful_handler.rb
```diff
@@ -91,27 +100,51 @@ def delete
...
+            current_logger.info(context: "Webhook check for alpha site successful", webhook_url: webhook.url)
           end
           if (webhook.name == "Publish to Usurper Beta (#{expected_release})") && (webhook.url == "https://wse-websiterenovation-libnd#{expected_release}-api.library.nd.edu/usurpercontent/entry")
             found_beta_webhook = true
-            current_logger.info(context: "Webhook set for beta site", webhook_url: webhook.url)
+            current_logger.info(context: "Webhook check for beta site successful", webhook_url: webhook.url)
+          end
+          if (webhook.name == "Libguides Event Updater (#{expected_release})") && (webhook.url == "https://wse-websiterenovation-#{expected_release}-api.library.nd.edu/monarchlibguides/newevent")
+            found_libguides_webhook = true
+            current_logger.info(context: "Webhook check for libguide event updater successful", webhook_url: webhook.url)
```
jeremyf: Should the following block on line 156 be added as an `else`? You could then get rid of `found_libguides_webhook` local variable.

spec/contentful/integration/int_contentful_spec.rb
h-parekh: Marking this as a TODO item, and tracking as a enhancement issue
https://github.com/ndlib/QA_tests/issues/233

spec/contentful/integration/int_contentful_spec.rb
h-parekh: Added

spec/spec_support/contentful_handler.rb
h-parekh: contentful_entry_id will be more beneficial, and thats being logged on line#55 after the entry is created.

spec/spec_support/contentful_handler.rb
h-parekh: I just came to know today that we have a new 'prep' contentful space that can be leveraged for testing purposes. So this was me getting a foot in the door to start using it. 
Definitely need to make this more robust. 

https://github.com/ndlib/QA_tests/issues/234

spec/spec_support/contentful_handler.rb
h-parekh: Yes, website will have broken functionality if any of these webhooks are missing,

spec/spec_support/contentful_handler.rb
h-parekh: I guess not, because these are not 'either-or' webhooks. All 3 have to be present in order for the site to fully function.

spec/spec_support/contentful_handler.rb
jeremyf: Is there an assertion enforcing this functionality being broken? As it looks, there would be a log entry for the error but that would not bubble up as a failed scenario.

h-parekh:
Yes the call to the API will fail if these webhooks are missing. At which point, these webhook pre-checks provide additional information to troubleshoot. 

I'm not explicitly asserting for these webhooks though.

PR #242, opened 2017-11-17
Changes lib_cal_id from hard code to dynamic

mhallwac:
Previously, the value for lib_cal_id, which represents the contentful parameter,
was hardcoded to a set value.  Now, the scenario gets the value for lib_cal_id
from the test_data/QA/parameters/contentful.yml file hen randomly picks the lib_cal_id
from the available options.  This adds flexibility for the values and helps avoid future code changes
if the values from the yml file change.

### DIFF
--- a/spec/contentful/integration/int_contentful_spec.rb
+++ b/spec/contentful/integration/int_contentful_spec.rb
@@ -23,10 +23,8 @@ feature 'Tests for Contentful entries and webhook integrations' do
   # I am calling /newevent via contentful to make sure the webhooks are firing
   # correctly, and the event is then previewable
   # 'lib_cal_id' is a unique ID that is used to pull data from the libcal system
-  # TODO: Parameterize libcal ID, we should leverage the PathParameterizer::ParameterFinder
-  # class. It probably needs some refactor becuase it's designed to parameterize
-  # endpoints for path and query parameters
-  # Tracking here: https://github.com/ndlib/QA_tests/issues/233
+  # libcal ID is parameterized by leveraging the PathParameterizer::ParameterFinder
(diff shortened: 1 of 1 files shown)


spec/contentful/integration/int_contentful_spec.rb
```diff
@@ -28,7 +28,9 @@
   # endpoints for path and query parameters
   # Tracking here: https://github.com/ndlib/QA_tests/issues/233
   scenario "Create and preview an entry of type 'Events'" do
-    ContentfulHandler.create(current_logger: current_logger, content_type: 'event', lib_cal_id: '3596276') do |entry|
+    contentful_params = YAML.load_file(ENV['HOME']+"/test_data/QA/parameters/contentful.yml")
```
h-parekh: Consider following [DRY principles](https://gist.github.com/harssh-sparkway/8707634), since we have existing code in the [ParameterFinder class](https://github.com/ndlib/QA_tests/blob/master/spec/spec_support/path_parameterizer.rb#L72) to load parameter files and build a dictionary.

spec/contentful/integration/int_contentful_spec.rb
```diff
@@ -28,7 +28,9 @@
...
+    lib_cal_id = contentful_param['libCalId'][0].values.sample.to_s
```
h-parekh: Consider moving this into a new method definition under [PathParameterizer module](https://github.com/ndlib/QA_tests/blob/master/spec/spec_support/path_parameterizer.rb#L3).

h-parekh: Also, as part of this pull request, we should update the comments for this scenario.

PR #245, opened 2018-01-03
Fixes curate tests failing due to failed network resources

h-parekh:
## Updates VerifyNetworkTraffic module

5127e24d536a0466634c93c87691a9222dd9e370

* Adds an array called 'EXCLUDE_URI_FROM_NETWORK_TRAFFIC_VALIDATION'
that can be used as a hook to exclude URIs in any spec
* Filters out non-URI resources because they don't have response codes
and their validation needs to be handled differently
* Skips verification of resources that have been included in
EXCLUDE_URI_FROM_NETWORK_TRAFFIC_VALIDATION
* Categorizes all network resources into 4 groups, for ease of
troubleshooting and logging
* Adds documentation about the 'verify_network_traffic' method

## Adding resource to be excluded for curate tests to pass

190ace602b86d989e9f2e3301d50848a669b392a

## Explicitly require Curate::Pages::HomePage

df3e6c91c256e8b11c6e68552a95220f26595150

For some reason, the curate_spec_helper won't require all files
in spec/curate/pages unless I first explicitly require
`spec/curate/pages/home_page.rb`

Without this change specs fail with an error:
``` console
uninitialized constant Curate::Pages::HomePage
```

### DIFF
--- a/spec/curate/pages/audio_page.rb
+++ b/spec/curate/pages/audio_page.rb
@@ -3,7 +3,7 @@ module Curate
     class AudioPage
       include Capybara::DSL
       include CapybaraErrorIntel::DSL
-
+      VerifyNetworkTraffic::EXCLUDE_URI_FROM_NETWORK_TRAFFIC_VALIDATION << '/assets/ui-bg_highlight-soft_100_eeeeee_1x100.png'
       def on_page?
           on_valid_url? &&
           status_response_ok? &&
--- a/spec/curate/pages/audio_page.rb
+++ b/spec/curate/pages/audio_page.rb
@@ -3,7 +3,7 @@ module Curate
     class AudioPage
       include Capybara::DSL
       include CapybaraErrorIntel::DSL
-      VerifyNetworkTraffic::EXCLUDE_URI_FROM_NETWORK_TRAFFIC_VALIDATION << '/assets/ui-bg_highlight-soft_100_eeeeee_1x100.png'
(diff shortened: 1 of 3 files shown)


spec/curate/pages/audio_page.rb
```diff
@@ -3,7 +3,7 @@ module Pages
     class AudioPage
       include Capybara::DSL
       include CapybaraErrorIntel::DSL
-
+      VerifyNetworkTraffic::EXCLUDE_URI_FROM_NETWORK_TRAFFIC_VALIDATION << '/assets/ui-bg_highlight-soft_100_eeeeee_1x100.png'
```
jeremyf: Normally, I don't like appending to a constant (that indicates a "non-constant"). To do this we'd need to create a class level instance variable. Based on the changes in `VerifyNetworkTraffic` you can make the following changes:

```ruby
VerifyNetworkTraffic.exclude_uri_from_network_traffic_validation += '/assets/ui-bg_highlight-soft_100_eeeeee_1x100.png'
```

spec/spec_support/test_runtime.rb
```diff
@@ -164,6 +164,7 @@ module VerifyNetworkTraffic
   # leverages network_traffic method of poltergeist driver to verify
   # that all the network calls for static assets and 3rd party support
   # are working on the site
+  EXCLUDE_URI_FROM_NETWORK_TRAFFIC_VALIDATION = []
```
jeremyf: Convert from constant to instance variable:

```ruby
def self.exclude_uri_from_network_traffic_validation
  @exclude_uri_from_network_traffic_validation ||= []
end
```

PR #263, opened 2018-01-22
Mh inquisition

mhallwac:
Commit 1
* Creates folder structure
* Creates pprd and prod urls for .config file
* Sets up inquisition_spec_helper.rb

Commit 2
* Creates the scenario for loading the homepage with smoke_test tag
* Creates the HomePage class to ensure the homepage loads with header, footer, and page content
* Updates spec_helper to require page files
* Updates func_inquisition_spec to use correct spec_helper call

### DIFF
--- a/spec/inquisition/inquisition_config.yml
+++ b/spec/inquisition/inquisition_config.yml
@@ -1,0 +1,2 @@
+pprd: https://inquisition-prep.library.nd.edu
+prod: https://inquisition.library.nd.edu
--- a/spec/inquisition/inquisition_config.yml
+++ b/spec/inquisition/inquisition_config.yml
@@ -1,2 +1,2 @@
-pprd: https://inquisition-prep.library.nd.edu
+prep: https://inquisition-prep.library.nd.edu
 prod: https://inquisition.library.nd.edu
--- a/spec/inquisition/pages/home_page.rb
+++ b/spec/inquisition/pages/home_page.rb
@@ -9,7 +9,6 @@ module Inquisition
       include CapybaraErrorIntel::DSL
 
       def on_page?
-        require 'byebug'; debugger
         status_response_ok? &&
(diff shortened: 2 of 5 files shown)


spec/inquisition/inquisition_config.yml
```diff
@@ -0,0 +1,2 @@
+pprd: https://inquisition-prep.library.nd.edu
```
h-parekh: Can you change the 'pprd' key to 'prep'. There was a collective discussion across the teams to make sure we all refer to the different environments in the same way.

spec/inquisition/pages/home_page.rb
```diff
@@ -0,0 +1,69 @@
...
+          has_facets? &&
+          has_title?
+      end
+
+      def status_response_ok?
+        status_code == 200 || status_code == 304
+      end
+
+      def on_valid_url?
+        current_url == Capybara.app_host || File.join(Capybara.app_host, '#')
```
h-parekh: This statement will always evaluate as true. 
You want to write it as:
```ruby 
current_url == Capybara.app_host || current_url == File.join(Capybara.app_host, '#')
```

spec/inquisition/pages/home_page.rb
```diff
@@ -0,0 +1,71 @@
...
+require 'inquisition/inquisition_spec_helper'
+
+module Inquisition
+  module Pages
+    class HomePage
+      include Capybara::DSL
+      include CapybaraErrorIntel::DSL
+
+      def on_page?
+        require 'byebug'; debugger
```
h-parekh: @mhallwac Remove the byebug statement, did rubocop identify this? I would think it should.

mhallwac: I just removed it, I think I ran byebug after rubocop

Issue #279, opened 2018-02-13
Add hook in a integration specs to specify its swagger definition


h-parekh:
### Descriptive summary
Currently, we are [reading swagger definitions](https://github.com/ndlib/QA_tests/blob/master/spec/spec_support/swagger_handler.rb#L47) for ALL integration tests from `definitions/swagger.yml` file of the respective APIs. We want to provide a hook within each APIs `spec/<app_name>/<app_name>_spec_helper.rb` to be able to tell it to use the API's `deploy/gateway.yml` file

### Rationale
We want to move away from using separate swagger.yml file for QA and use the deployment file itself. This saves us time and effort in communication between dev and QA cycles.

### Done looks like
* Add a method in SwaggerHandler module to read from swagger definition from deploy/gateway.yml
* Update `spec/monarch_libguides` to toggle between using `definitions/swagger.yml` and `deploy/gateway.yml`
* Update README file for instructions

h-parekh:
Addressed by https://github.com/ndlib/QA_tests/pull/280

PR #280, opened 2018-02-16
Creates swagger_from_gateway method and creates toggle hook

mhallwac:
* Based on https://github.com/ndlib/QA_tests/issues/279
* The SwaggerHandler method, swagger, has been renamed swagger_from_definitions because definitions/swagger.yml is the location of that swagger file
  * Code has not been changed
* Creates the new method swagger_from_gateway which is similar to swagger_from definitions except it gets the swagger definition from deploy/gateway
  * Removed initial attempt to build swagger and rescue ArguementError because swagger definition is always nested in gateway files
* Adds require_swagger_location which forces tester to specify where the swagger should be loaded from in an ENV hook
  * ex: SWAGGER_LOCATION=gateway
  * SwaggerHandler.require_swagger_location is run in spec/spec_helper in the before(:suite) hook because SwaggerHandler is instantiated before examples
* Adds ensure_integration_test method to SwaggerHandler to ensure that SWAGGER_LOCATION is only required for integration tests

### DIFF
--- a/spec/spec_support/swagger_handler.rb
+++ b/spec/spec_support/swagger_handler.rb
@@ -26,12 +26,20 @@ class SwaggerHandler
   # Sets @operations variable with an array of objects of type SwaggerHandler::SwaggerOperationDecorator
   def operations
     @fitered_operations = reject_operations_with_skip_tag
-    @operations ||= @fitered_operations.map { |operation| SwaggerOperationDecorator.new(swagger, operation) }
+    if ENV['SWAGGER_LOCATION'] == 'gateway'
+      @operations ||= @fitered_operations.map { |operation| SwaggerOperationDecorator.new(swagger_from_gateway, operation) }
+    else
+      swagger_from_definitions.operations.delete_if { |operation| !operation.tags.nil? && operation.tags.include?("skipTests") }
+    end
   end
 
   # Returns an array of objects of type Swagger::V2::Operation
   def reject_operations_with_skip_tag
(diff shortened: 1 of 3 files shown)


spec/spec_support/swagger_handler.rb
```diff
@@ -72,6 +90,23 @@ def set_access_token_value!
       @access_token_value = git_access_config_file.fetch("access_token")
     end
 
+    def self.require_swagger_location
+      if SwaggerHandler.ensure_integration_test
+        if ENV['SWAGGER_LOCATION'].nil?
+          Bunyan.current_logger.error(context: "SWAGGER LOCATION not found in ENV config")
```
h-parekh: Convey the actual variable name you want the user to use for passing in ENV
```ruby
Bunyan.current_logger.error(context: "SWAGGER_LOCATION not found in ENV config")
Bunyan.current_logger.error(context: "Provide SWAGGER_LOCATION of API being tested, ex: SWAGGER_LOCATION=gateway")
```

mhallwac: SWAGGER_LOCATION is the variable name, I followed the same format as require_version_number  

~~~~
def self.require_release_number
    if ENV['RELEASE_NUMBER'].nil?
      Bunyan.current_logger.error(context: "RELEASE_NUMBER not found in ENV config")
      Bunyan.current_logger.error(context: "Provide RELEASE_NUMBER of application being tested, ex: r20170922")
      exit!
   end
~~~~

PR #281, opened 2018-02-19
Usurper website release r20180216

h-parekh:
* Search toggle has been removed from banner


h-parekh:
Yes new release is visible on alpha.library.nd.edu and beta.library.nd.edu

PR #284, opened 2018-02-27
Add specs for file browsing/pagination

jgondron:
## DLTP-1261: Add specs for file browsing/pagination

01ea4980fd08eb60675e7498901b7b837384eac5

- Added a way to disable maleficent. This is useful when iterating on a test that you know is failing
- Added a method to allow searching for a term with spaces. When this happens, the url of the page uses an encoded version of the search term. Changed on_valid_url? to check for encoded search term in the url
- Added a helper method within Article to test if pagination was rendered for the article and with the correct params
- The existing show article spec was not testing file downloads. This has been silently failing for a few reasons. First, the first article that it selects in the search may or may not have files (currently does not for pprd, for example). Second, the has_files? check was silently failing due to ambiguity if the article had multiple files. Fixed the ambiguity, moved this into a feature spec for testing file browsing and added specs for testing that pagination is rendered for an article with many files. Also fixed a problem with how it tests that the file was downloaded. Testing page url is not valid, have to check the response headers.

### DIFF
--- a/spec/curate/functional/func_curate_spec.rb
+++ b/spec/curate/functional/func_curate_spec.rb
@@ -99,15 +99,6 @@ feature 'User Browsing', js: true do
     show_article = Curate::Pages::ShowArticlePage.new(title: article_link.text)
     article_link.click
     expect(show_article).to be_on_page
-    if show_article.has_files?
-      download_button=find("a.action.btn", :text => "Download")
-      article_link = download_button['href']
-      download_button.click
-      last_opened_window = page.driver.browser.window_handles.last
-      page.driver.browser.switch_to_window(last_opened_window)
-      expect(page.current_url).to eq(article_link)
-      expect(status_code.to_s).to match(/^20[0,1,6]$/)
-    end
   end
 end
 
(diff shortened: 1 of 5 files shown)


spec/curate/pages/catalog_page.rb
```diff
@@ -65,7 +65,7 @@ def on_valid_url?
           current_url.start_with?(File.join(Capybara.app_host, 'catalog'))
         elsif category.nil?
           current_url.start_with?(File.join(Capybara.app_host, 'catalog')) &&
-            current_url.include?(search_term)
+            current_url.include?(encoded_search_term)
```
h-parekh: Is this change going to enforce encoding when the search term doesn't have spaces?

spec/curate/functional/func_curate_spec.rb
```diff
@@ -548,3 +539,36 @@
     category_page.test_facets_list_vs_grid
   end
 end
+
+feature 'Browsing attached files' do
+  scenario "Show an Article with many files" do
+    visit '/'
+    home_page = Curate::Pages::HomePage.new
+    expect(home_page).to be_on_page
+    search_term = "Article with many files"
```
h-parekh: My understanding is that this scenario will only work on a result with more than 10 attached files for pagination to be activated. Correct?

h-parekh:
* Very thoughtful in adding SKIP_MALEFICENT toggle, love it!
* Added some comments inline

h-parekh:
I'm also thinking if results with 'Type of work=Books' will almost always have attached files. But that may not give us a page with more than 10 attached files which is what we need to test for pagination.

spec/curate/pages/catalog_page.rb
jgondron: Yes, but so far that hasn't caused any issues with the existing specs. It could cause a mismatch down the line though. I figured we'd just have to deal with that once we have more examples.

spec/curate/functional/func_curate_spec.rb
jgondron: Yes. This is why I'm searching for a particular article. This is the part we need to discuss, since this article may or may not exist in the target environment.

PR #286, opened 2018-03-15
Hp tags

h-parekh:
## Adds SpecFilterManager module

2a0ab09a1f4d3a72c9f3abf4ecc908d2971c48d0

* Adds SpecFilterManager module
 This module will provide spec filtering capabilities to the framework
 based on value of ENVIRONMENT variable.
* Adds 'set_default_filters' method to the module
 If the ENVIRONMENT variable is a valid key in the spec's config file
 then it proceeds to determine if its a 'production' or 'non_production'
 based on these values:
 possible_production_keys = ['prod']
 possible_non_production_keys = ['local', 'dev', 'test', 'prep', 'pprd', 'staging', 'staging6', 'staging8','staging9']
* For 'production' endpoints, this method configures Rspec to include
only scenarios that are tagged as 'read_only' and explicitly excludes
scenarios tagged as 'nonprod_only' (to handle scenarios that have both these tags)
In that case Rspec will not run the scenario
 @example The following scenario will not run:
 ```ruby
 scenario 'Loads Home Page',:read_only, :nonprod_only do
   visit '/'
 end
 ```
 * I'm also explicitly requiring the user to input 'ENVIRONMENT_CATEGORY'
 in case they run the spec with the 'ENVIRONMENT' value as a http URL.
 I'm trying to avoid making any assumptions about whether the URL is
 considered 'production' or 'non_production'. Especially in cases when
 URL like: http://dld625i.cloudfront.net or https://5sm5qzp.execute-api.us-east-1.amazonaws.com/r20180302

## Adding 'read_only' tags to all scenarios

5a75f311cc9cf11896393afd4de8fb0c92f8605a

## Updates 'bin/run_tests' and README file

531234983d150e24b9fd7626090b0b5432a567bd

* Adds support for accepting 'ENVIRONMENT_CATEGORY' variable in 'run_tests'
* Consolidates test running instructions in 'bin/run_tests' to avoid
duplication and inconsistencies

## Pending changes
* Add support for Jenkins to be able to handle these changes. I need this PR to merged before I can test with Jenkins
* Customize Rubocop to enforce tagging of each scenario

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,10 +173,11 @@ $ ./bin/run_tests spec/curate/functional/func_curate_spec.rb
 
 #### Testing on your local machine
 
+To see all available input variables and examples:
 ```console
 $ cd /path/to/QA_tests
-$ ./bin/run_tests spec/curate/functional/func_curate_spec.rb
-```
+$ ./bin/run_tests -h
+````
 
 #### Triggering from Jenkins CI
 
--- a/bin/run_tests
+++ b/bin/run_tests
@@ -11,9 +11,10 @@ end
 #
 # *****************************************************************************
 
-CONFIG_KEYS = [:ENVIRONMENT, :LOG_LEVEL, :SKIP_CLOUDWATCH, :SKIP_VERIFY_NETWORK_TRAFFIC, :SKIP_MALEFICENT, :ALLOW_ALL_NETWORK_HOSTS, :USE_CONTENTFUL_SPACE, :VERSION_NUMBER, :RELEASE_NUMBER].freeze
+CONFIG_KEYS = [:ENVIRONMENT, :ENVIRONMENT_CATEGORY, :LOG_LEVEL, :SKIP_CLOUDWATCH, :SKIP_VERIFY_NETWORK_TRAFFIC, :SKIP_MALEFICENT, :ALLOW_ALL_NETWORK_HOSTS, :USE_CONTENTFUL_SPACE, :VERSION_NUMBER, :RELEASE_NUMBER].freeze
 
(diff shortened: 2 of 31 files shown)


jgondron:
Can you confirm that this will still allow us to use a focus tag on a spec? Ie, add a `focus: true` to a scenario, then on the command line run with `--tag focus`.

h-parekh:
@jgondron Yes, any command line tags will override Rspec.configuration. 

CAUTION: Note scenario 5 and example 3 below, even if Rspec.configuration is set to exclude `nonprod_only`, Rspec command line tag `focus` tag overrides that exclusion. 

```ruby
# spec/sample_tests.rb
scenario 1, :read_only do ... end
scenario 2, :nonprod_only do ... end 
scenario 3, :focus, :read_only do ... end
scenario 4, :focus, :nonprod_only do ... end
scenario 5, :focus, :read_only, :nonprod_only do ... end 
```

```sh
# Example 1
$ ENVIRONMENT=prod bin/run_tests spec/sample_tests.rb
# Will run scenarios 1 and 3

# Example 2
$ ENVIRONMENT=prep bin/run_tests spec/sample_tests.rb
# Will run scenarios 1,2,3,4,5

# Example 3
$ ENVIRONMENT=prod bin/run_tests spec/sample_tests.rb --tag focus 
# Will run scenarios 3, 4 and 5

# Example 4
$ ENVIRONMENT=prep bin/run_tests spec/sample_tests.rb --tag focus 
# Will run scenarios 3,4 and 5
```

h-parekh:
@jgondron Did the previous comment help clarify your question?

PR #294, opened 2018-04-12
Add tests for inquisitions error pages

jgondron:
## QA-46, DLTP-1345: Add tests for inquisitions error pages

a357ea120c3cba24c116b37ed48e2882da30a4cf

- Added an ErrorPage spec helper and added a few known scenarios where the app should render these custom error pages
- Changed spec_helper to clear the VerifyNetworkTraffic.exclude_uri_from_network_traffic_validation on tear down of each test. Adding urls to this list will now need to be done as part of each scenario that cares about it, not just when the modules are loaded. In all cases I found, we're using Page helpers to construct this list, so added it to their initialize methods. Currently this will impact primo, inquisition, curate, and usurper tests.
- This does not cover pages that should load with 200

### DIFF
--- a/spec/curate/pages/audio_page.rb
+++ b/spec/curate/pages/audio_page.rb
@@ -3,7 +3,11 @@ module Curate
     class AudioPage
       include Capybara::DSL
       include CapybaraErrorIntel::DSL
-      VerifyNetworkTraffic.exclude_uri_from_network_traffic_validation.push('/assets/ui-bg_highlight-soft_100_eeeeee_1x100.png')
+
+      def initialize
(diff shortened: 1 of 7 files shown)


h-parekh:
I've verified these specs with/without your changes and they're working as expected:
```console
SKIP_CLOUDWATCH=true ENVIRONMENT=prep bx bin/run_tests spec/inquisition/functional/func_inquisition_spec.rb
VERSION_NUMBER=v2018.10 SKIP_CLOUDWATCH=true bx bin/run_tests spec/usurper/functional/func_usurper_spec.rb:32
SKIP_CLOUDWATCH=true bx bin/run_tests spec/curate/functional/func_curate_spec.rb:433
ENVIRONMENT=prep SKIP_CLOUDWATCH=true bx bin/run_tests spec/primo/functional/func_primo_spec.rb:6

PR #307, opened 2018-05-21
Hp docker on aws

h-parekh:
## Updates docker-compose

16d89502af8ed82aeb2438d120625174c8da7176

* Adds a new service in docker-compose for AWS
* Removes dependency on corpFS from docker-compose.yml
* Adds '-dev' suffix to make the service descriptive

## Upgrades aws-sdk gem from v2 to v3

5c401173481135642e7185b6d6836054bd1d5d95

## Includes Gemfile and .rspec in the docker image

50f74b629698ad5c91902b52fd207e7250e36131

## Adds a sample template to trigger tasks in AWS ECS

41c78fb0066ee909340e1290a43b84721bb4690e

## Adds AwsSsmHandler module for creating SSM clients

5f5a8941af67ca61a955217b2fde02177ad65d26

* This commit creates a new AwsSsmHandler module which is designed to
handle two scenarios:
1. Running on local, (RUNNING_ON_LOCAL_DEV=true)
In this scenario, the ssm client is created with an assumed IAM role.
This is necessary because secrets in AWS parameter store access
can only be read by a specific IAM role. A developer would not be
able to read them locally using their own AWS account

2. Running on AWS ECS, (RUNNING_ON_LOCAL_DEV=false)
In this scenario, the ssm client is created with an instance profile
credentials object that can be accessed by leveraging
Aws::InstanceProfileCredentials. This object contains credentials
of the EC2 instance in which the docker service is running in AWS, which
has permissions setup to read from the AWS parameter store

* This commit also exposes a `get_param_from_parameter_store` method
that can be used across other modules to read secrets from the parameter
store

## Enables reading secrets from parameter store

a2e7a9cfc557807c3ccb70236f6546970f9db42b

In this commit I am using AwsSsmHandler.get_param_from_parameter_store
method to consume secrets from AWS parameter Store.
Since the secrets have been converted from yaml to json structures
there are also some changes in logic to read specific values
* LoginPage.initialize method reads secrets from
"/all/qa/testing_netids"
* ContentfulHandler.read_contentful_tokens method reads secrets
from "/all/qa/cf_api_key"
* ParameterFinder.build_dictionary! method reads secrets from
"/all/qa/shared" and "/all/qa/#{application_name_under_test}"
* Adds sampling logic on subdictionary. The ParameterFinder.find
method is used in 2 use cases:
1. From InitializeExample.find_url_from_secret_files for looking up
value to set Capybara.app_host by key (prep, prod.. ). In this
case its a 1:1 key:value on a Hash object
2. From PathParameterizer.call and
PathParameterizer.call_query_parameterizer for looking up values of
testing data. In this case its a 1:many lookup on a Array of values
from which I want to pick 1 testing value
I don't expect the dictionary to have any other form of data, and so
I'm raising an error for anything other than Hash/Array
* RequestBuilder.get_token method reads JWT token from
"/all/qa/jwt_token"

## Disables rspec result reporting to cloudwatch

633f4c06590dd286d383f1f190fceddb93aa7db6

This approach is broken in its current state, and I want to revisit
metrics reporting. So I'm disabling reporting with this commit instead
of putting effort that will be thrown away eventually.

Rspec results and logs are still accessible in the ECS task logs in
cloudwatch

## Updates README

7458eb2b1aa8ab8f21db39e7151355951bc9f48b

* Removes documentation about running on testcontroller and corpfs
* Updates setup information for AWS cli
* Adds examples on how to run tests using the various env variables
on local host as well as in AWS


docker-compose.yml
```diff
@@ -1,12 +1,13 @@
 ---
 version: '3'
 services:
-  qa-tests:
+  qa-tests-dev:
     build: .
-    volumes:
-    - /Volumes/DCNS/Library/Departmental/Infrastructure/vars/QA:/root/test_data/QA
+    env_file:
+      - ~/.aws/credentials
```
jgondron: You may want to rethink using these files as env. They are not conducive to being used as env, even if they are key value pairs (which is why you're having to do things like https://github.com/ndlib/QA_tests/pull/307/files#diff-d05495e4b9bb346fbb9bfd5efe2f50e3R10). You're also making an assumption that there is only one profile defined. Several of us will have multiples defined in these at this point.

Also something to consider (at least when running locally) is that several of us use a script to assume a role for a prolonged session. Using this makes it easy to assume a role, enter our mfa token, and not have to do this again for a while. This script works by setting a few known env variables (see https://docs.aws.amazon.com/cli/latest/userguide/cli-environment.html). So perhaps for -dev env we can just read these straight from env instead of a file? That or we can change it to pipe the output from this scrip to a .env. Either could work. If we did this, this https://github.com/ndlib/QA_tests/pull/307/files#diff-d05495e4b9bb346fbb9bfd5efe2f50e3R43 could probably be skipped if the AWS_SESSION_TOKEN is defined, since this would mean the user has already assumed a role themselves.

Might be easier to just chat about this...

h-parekh: My chain of thought is to supply credentials to the docker container as environment variables. So I'm providing [instructions](https://github.com/ndlib/QA_tests/tree/hp_docker_on_aws#setting-up-aws-profiles) on how to setup the correct profile and export it before starting to run tests. 

Agree that this implementation falls short because I still have to provide a token code everytime, curious about the script you're referring to. Perhaps I should have something like that. Can you provide a link to that script?

PR #342, opened 2019-03-01
Cleanup bendo spec

ialford:
- Moved the Ruby Bendo spec to a /retired/ path, but kept the files intact in case they are needed.

- Added a README to link to both the location of the retired Ruby specs as well as the new Newman specs.


ialford:
> It looks good, can we also add a sentence explaining why we chose to move...eg: to consolidate them in single code and store them in lieu with their service(bendo)

Can you provide a review of the updated README and let me know if this meets your concerns?

rmangu:
Perfect , Thanks Ian !

PR #344, opened 2019-09-24
selenium script for homepage tests

ialford:
This is the first in, likely, a long series of updates as we rework the website QA stack.

I will be following up with documentation on how to run and possibly scripting around that as well.


h-parekh:
Glad to see this repo getting a new life!

PR #347, opened 2019-09-27
initial QA specs and instructions for Aleph

ialford:
Creating an initial suite of some basic API and OPAC checks, along with instructions to follow.

This should close out ESU-973 and future enhancements can be done in other tickets.

### DIFF
--- a/spec/aleph/newman/Aleph.postman_collection.json
+++ b/spec/aleph/newman/Aleph.postman_collection.json
@@ -334,7 +334,7 @@
 							"    pm.expect(pm.response.responseTime).to.be.below(1000);",
 							"});",
 							"pm.test(\"Body matches string\", function () {",
-							"    pm.expect(pm.response.text()).to.include(\"LF 509\");",
+							"    pm.expect(pm.response.text()).to.include(\"<div class=\\\"library_college\\\">University of Notre Dame</div>\");",
 							"});"
 						],
 						"type": "text/javascript"
@@ -345,14 +345,14 @@
 				"method": "GET",
 				"header": [],
 				"url": {
-					"raw": "https://{{AlephURL}}/F/1Q3EMSL9NCKB6P19QU5QRKMCY4QNLBDCUTADR3LJS7YL33JEK5-08833?func=direct&doc_number=000123456",
+					"raw": "https://{{AlephURL}}/F/{{$guid}}?func=direct&doc_number=000123456",
 					"protocol": "https",
 					"host": [
 						"{{AlephURL}}"
 					],
 					"path": [
 						"F",
-						"1Q3EMSL9NCKB6P19QU5QRKMCY4QNLBDCUTADR3LJS7YL33JEK5-08833"
+						"{{$guid}}"
 					],
 					"query": [
 						{
--- a/spec/aleph/newman/Aleph.postman_collection.json
+++ b/spec/aleph/newman/Aleph.postman_collection.json
@@ -1,0 +1,373 @@
+{
+	"info": {
+		"_postman_id": "643f1b11-08eb-4855-9761-eae16bbed0c0",
+		"name": "Aleph",
+		"description": "## QA - Aleph API Calls\n\nTests designed to make calls against the Aleph API to determine functionality of system. Additional calls can be added later once need is discovered.",
+		"schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json"
+	},
+	"item": [
+		{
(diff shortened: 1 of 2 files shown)


ialford:
> Works for alephprod. Now we just have to figure out what is missing on the new Aleph23 servers. They don't yet pass.

There's more success if you add the `--insecure` flag after `Aleph.postman_collection.json` in the CLI. The RESTful and OPAC-Search tests themselves fail, but this could be a configuration or data issue.

spec/aleph/newman/Aleph.postman_collection.json
```diff
@@ -0,0 +1,373 @@
...
+						"id": "09f59499-3342-49ee-aa5a-2bf76213ee79",
+						"exec": [
+							"pm.test(\"Status code is 200\", function () {",
+							"    pm.response.to.have.status(200);",
+							"});",
+							"pm.test(\"Response time is less than 1000ms\", function () {",
+							"    pm.expect(pm.response.responseTime).to.be.below(1000);",
+							"});",
+							"pm.test(\"Body matches string\", function () {",
+							"    pm.expect(pm.response.text()).to.include(\"Item could not been found based on item barcode:\");",
```
abbales: Looks like a type "been found". Is this the string as Aleph returns it?

spec/aleph/newman/Aleph.postman_collection.json
```diff
@@ -0,0 +1,373 @@
...
+						],
+						"type": "text/javascript"
+					}
+				}
+			],
+			"request": {
+				"method": "GET",
+				"header": [],
+				"url": {
+					"raw": "https://{{AlephURL}}/F/1Q3EMSL9NCKB6P19QU5QRKMCY4QNLBDCUTADR3LJS7YL33JEK5-08833?func=direct&doc_number=000123456",
```
abbales: Looks like the URL has a session ID in it (the string following F/). You might need to eliminate this, or if a session ID is needed to connect with an earlier call, should it use a variable?

spec/aleph/newman/Aleph.postman_collection.json
```diff
@@ -0,0 +1,373 @@
...
+					"protocol": "https",
+					"host": [
+						"{{AlephURL}}"
+					],
+					"path": [
+						"F",
+						"1Q3EMSL9NCKB6P19QU5QRKMCY4QNLBDCUTADR3LJS7YL33JEK5-08833"
```
abbales: Again, looks like a session ID.

spec/aleph/newman/Aleph.postman_collection.json
ialford: Yes, it is, @abbales . Aleph returns `Item could not been found based on item barcode` if you send in a bad barcode, which we are doing by virtue of sending in a randomly generated GUID.

spec/aleph/newman/Aleph.postman_collection.json
ialford: We can make those changes if needed - is there something in the session ID that would cause the tests to fail or succeed unexpectedly? Such as moving between systems (such as from `alephprod` to `alephpprd`?

spec/aleph/newman/Aleph.postman_collection.json
ialford: See previous comment

spec/aleph/newman/Aleph.postman_collection.json
abbales: OK. I just wanted to be sure the typo was in Aleph and not a mistake in the code. Thanks!

spec/aleph/newman/Aleph.postman_collection.json
abbales: The session ID is randomly generated when you start a web interaction, and then they expire. I don't remember how long they last -- perhaps 24 or 48 hrs, or something. I'm not sure what will happen if you pass a session ID that no longer exists with an API call.

ialford: @abbales Do you think an adequate substitution here would be to make a request with a bad session ID and test against the expected results of that?

We could probably also specifically request a Session ID and use it later, but that would be more involved. Thoughts?

ialford: Re-wrote the spec and requested a new review from @abbales

ialford:
I'm going to merge this as it is non-destructive. I will follow up with Aaron later.

PR #348, opened 2019-10-14
update one of the response speeds to response code

ialford:
There was a doubling of response time causing false negative. This fixes the 200ms time to a 200 response code.


ialford:
> So, rather than trying to get a certain response time level this switches to just making sure it is running? Was this because it could not consistently meet that time or because we just don't want to put such requirements into place?

The previous code had two response time checks - one for `200ms` and one for `1000ms` (which is the one that we care about). We are removing the `200ms` check in favor of a `200` response code - therefore the test will pass if we get a `200 - OK` within one second from the server.

With the original code, we were getting false failures - the server would respond `200 - OK` within `250ms` but the tests would fail, because it wanted the time to be under `200ms`. Does that make sense @hanstra ?

hanstra:
Yep, makes perfect sense. Aleph *is* slow so it does not surprise me that it times out regularly

PR #363, opened 2020-04-24
Esu 1264

wsill:
This is for a basic smoke test of some of the EC2 servers and VMware servers I created. Eventually I plan on adding additional test as needed.

As you can see I created an ec2 and vmware folder and then added subfolders based on the FQDN of the server I test.  The reason I did it this way was it allowed me to setup a qa-ec2 and qa-vmware project on lib-jenkins and dynamically pick from a list which server I need to test rather than setting an individual project for each server.  See the https://lib-jenkins-test.library.nd.edu/job/ec2-qa/ or https://lib-jenkins-test.library.nd.edu/job/vmware-qa/ project.

I'm looking for feedback to whether or not this (/spec/ec2/FQDN or /spec/vmware/FQDN) might be a better way to approach and organize our VM tests rather than the current (/spec/name) way it is being done now.


wsill:
> Because so many of these are the same with just the URL being different, if the plan is to go the route of drop-down menus in a single Jenkins job, a good consolidation may be to use a single base test and pass in a variable for the URL based on the Jenkins run.
> 
> There also does not appear to be any actual tests - just the GET calls. You can load up one of the JSON files from another repo, like [apache-blueprints](https://github.com/ndlib/apache-blueprints/blob/master/spec/newman/Apache_Website.postman_collection.json) to see an example of including tests in the calls. Look for the "Tests" section in Postman - there are even some pre-written tests Postman provides that would be a good starting point.

I put those json files in there so I could test, they don't really represent what eventual tests might end up being.  However, I did add an actual test for suma so I have one in there now. 

Many of the test will only be a simple test of the base URL response which of course I could put in as a simple base test. But, what I'm struggling with is deciding whether I should make the drop down list dynamic (which means some way to figure that out such as tag values, folder names, etc.) or having the drop down list be a manual change in the project.  I'm also struggling with do I configure the project as individual tests (test Apache, test IIS, test API, etc.) or do I attack it as individual servers (suma, jenkins, etc).  I'm hoping it will become more clear to me as I go further down this smoke test path.

Thanks for the feedback.

PR #370, opened 2021-05-07
feat: add puppeteer testing for e2e login checks

ialford:
Due to the nature of the way SPAs are rendered in the browser, selenium doesn't work very well as a testing framework. This adds tests written in Puppeteer, which is a Node library and allows promises to be used as we wait for content to be rendered in the browser, making the front end tests more reliable. 

The tests are also written in a jest style format, making them as approachable as unit tests for some. 

Here, we add two basic tests for expected content - that the browser is going to the right place and that a login button that works is present - along with an end-to-end login check, if appropriate ENV is leveraged. Instructions are in the README for both the Docker build/test and running the test suite without Docker.

Closes ESU-1142

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ logs/*.log
 vendor/*
 .byebug_history
 .vscode
+node_modules/
--- a/spec/usurper/puppeteer/Dockerfile
+++ b/spec/usurper/puppeteer/Dockerfile
@@ -1,0 +1,48 @@
+FROM node:14-slim
+
+# Install latest chrome dev package and fonts to support major charsets (Chinese, Japanese, Arabic, Hebrew, Thai and a few others)
+# Note: this installs the necessary libs to make the bundled version of Chromium that Puppeteer
+# installs, work.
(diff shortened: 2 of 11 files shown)


wsill (changes requested):
Will you need to update some of the packages so that all the tests pass?

ialford:
> Will you need to update some of the packages so that all the tests pass?

I can add updates to the Chrome container, sure. Would you prefer that happen in this PR or that it be addressed in a subsequent PR? @wsill

wsill:
> > Will you need to update some of the packages so that all the tests pass?
> 
> I can add updates to the Chrome container, sure. Would you prefer that happen in this PR or that it be addressed in a subsequent PR? @wsill



> > Will you need to update some of the packages so that all the tests pass?
> 
> I can add updates to the Chrome container, sure. Would you prefer that happen in this PR or that it be addressed in a subsequent PR? @wsill

Normally I would say in this PR but it might be better in a separate PR.

jrittenh (approved):

PR #375, opened 2021-05-21
feat: migrate convocate tests to puppeteer for chrome

ialford:
These tests as written merely replicate the tests that were written in Ruby/rspec previously. Once approved, the Ruby specs could be retired.

Closes ESU-993

### DIFF
--- a/spec/vatican/puppeteer/chrome/spec/e2e.test.js
+++ b/spec/vatican/puppeteer/chrome/spec/e2e.test.js
@@ -1,0 +1,223 @@
+const puppeteer = require("puppeteer");
+const url = process.env.BaseURL;
+
+describe("Convocate End to End Tests", () => {
+  let browser;
+  let page;
+
+  beforeAll(async () => {
+    browser = await puppeteer.launch({
+      args: ["--disable-dev-shm-usage"],
+    });
+  });
+
+  //Run each test on its own page with a default timeout of 10s
+  beforeEach(async () => {
+    page = await browser.newPage();
+    jest.setTimeout(10000);
+  });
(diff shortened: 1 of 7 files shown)


spec/vatican/puppeteer/chrome/spec/e2e.test.js
```diff
@@ -0,0 +1,223 @@
...
+  beforeAll(async () => {
+    browser = await puppeteer.launch({
+      args: ["--disable-dev-shm-usage"],
+    });
+  });
+
+  //Run each test on its own page with a default timeout of 10s
+  beforeEach(async () => {
+    page = await browser.newPage();
+    jest.setTimeout(10000);
```
jgondron, line 17: Everything seems fine, I would just be cautious of a blanket timeout that isn't a hard timeout (like 30s in the case of APIs, or longer in the case of sites). In most cases, we aren't testing performance here in these types of tests. If for some reason you just have a high latency for one execution, it can fail an entire build, which may be a false positive. Performance should generally be measured among a distribution of requests instead of just one request like this.

jgondron (approved):

spec/vatican/puppeteer/chrome/spec/e2e.test.js
ialford: Maybe 30 seconds, based on https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/RequestAndResponseBehaviorCustomOrigin.html#request-custom-request-timeout ?

jrittenh (approved):

